PR #8, opened 2019-11-07
Add datagram tag bits to RX packet message

JayKickliter:
Routers need the original golay-decoded tag bits to verify datagram fingerprints. This PR adds them in the most unobtrusive, if not pretty, way. 

Ideally, in a future PR, we'll make `LongFiRxPacket` contain a `one_of datagram`. It's not needed now though since we're only using `monolithic` datagrams.

### DIFF
--- a/src/hotspot_protocol/longfi.proto
+++ b/src/hotspot_protocol/longfi.proto
@@ -44,14 +44,17 @@ message LongFiRxPacket {
     uint32 oui = 5;
     // Device ID
     uint32 device_id = 6;
-    // Fingerprint 
+    // Fingerprint
     uint32 fingerprint = 7;
-    // Sequence 
+    // Sequence
     uint32 sequence = 9;
     // Spreading to be used
     LongFiSpreading spreading = 10;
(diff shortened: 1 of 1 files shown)


src/hotspot_protocol/longfi.proto
```diff
@@ -44,14 +44,17 @@ message LongFiRxPacket {
...
-    // Sequence 
+    // Sequence
     uint32 sequence = 9;
     // Spreading to be used
     LongFiSpreading spreading = 10;
     // the fully reassembled payload
     bytes payload = 11;
+    // De-golayed datagram id and flag bits.
+    // NOTE: only the lowest 12 bits are valid.
+    uint32 tag_bits = 12;
```
macpie: @Vagabond We should not have to update miner/router because of this right?

madninja: Additional fields would be ignored if router has a proto without that last one

Vagabond: I also need to use this in the router, so....

PR #9, opened 2019-11-11
Add simple rust library wrapper based on prost

madninja:


JayKickliter:
I've long wanted to have the `hotspot_protocol` protos wrapped in this repo also. If we were to do that, would they live in the same crate?

madninja:
> I've long wanted to have the `hotspot_protocol` protos wrapped in this repo also. If we were to do that, would they live in the same crate?

I think that;s fine? The crate library lib.rs name-spaces the various subsystems.. Would that work?

JayKickliter:
I think so, as long as it's not a breaking change for users of this code when the hotspot protos are wrapped and added.

Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+version = "0.1.0"
+authors = ["Helium <info@helium.com>"]
+edition = "2018"
+
+[dependencies]
+bytes = "0.4"
+prost = "0.5"
+
+[build-dependencies]
+prost-build = {version = "0.5"}
```
JayKickliter:
```suggestion
prost-build = {version = "0.5"}

```

src/blockchain/transactions/v1/txn_token_burn_v1.proto
```diff
@@ -11,4 +12,4 @@ message txn_token_burn_v1 {
     uint64 amount = 4;
     uint64 nonce = 5;
     bytes signature = 6;
-}  
\ No newline at end of file
+}
```
JayKickliter:
```suggestion
}

```

src/blockchain/transactions/v1/txn_rewards_v1.proto
```diff
@@ -13,11 +14,11 @@ message txn_reward_v1 {
...
+
 }
 
 message txn_rewards_v1 {
     uint64 start_epoch = 1;
     uint64 end_epoch = 2;
     repeated txn_reward_v1 rewards = 3;
-}   
\ No newline at end of file
+}
```
JayKickliter:
```suggestion
}

```

PR #19, opened 2020-02-13
add nonce to create_htlc txns

andymck:
This adds a nonce to the `blockchain_txn_create_htlc_v1` txn type and is to enable absorbed/2 checks in blockchain core.  The related blockchain core PR is:

https://github.com/helium/blockchain-core/pull/305


Vagabond:
I think we can take this out of draft status and merge it.

PR #21, opened 2020-03-09
update block format for performance voting

evanmcc:
proto changes for "voting"


src/blockchain_block_v1.proto
```diff
@@ -17,4 +17,6 @@ message blockchain_block_v1 {
     uint64 election_epoch = 7;
     uint64 epoch_start = 8;
     bytes rescue_signature = 9;
+    repeated bytes seen_votes = 10;
```
Vagabond: Do we want to track who voted for what here?

evanmcc: I don't know.  I thought about it.  it's not incredibly hard to add.

PR #23, opened 2020-03-26
Attach more time information to helium packets

Vagabond:


src/packet.proto
```diff
@@ -2,6 +2,23 @@ syntax = "proto3";
...
+    int32 altitude = 3; // meters?
+    uint32 time_accuracy = 4; //nanoseconds
+}
+
+// https://www.eecis.udel.edu/~mills/ntp/html/ntpq.html#system
+message ntp_info {
+    bool synced = 1;
+    uint32 stratum = 2;
+    float sys_jitter = 3; //combined system jitter, in milliseconds
+    float clock_jitter = 4; //clock jitter, in milliseconds
```
madninja: Do these have to be floats? vs some other fixed point representation?

src/packet.proto
```diff
@@ -2,6 +2,23 @@ syntax = "proto3";
 
 package helium;
 
+message gps_info {
+    uint32 horizontal_accuracy = 1; // meters? this will take into account h3 precision and distance from asserted location
+    uint32 vertical_accuracy = 2; // meters?
+    int32 altitude = 3; // meters?
+    uint32 time_accuracy = 4; //nanoseconds
```
madninja: Should this be uint64?

src/packet.proto
```diff
@@ -10,9 +27,14 @@ message packet {
     }
     packet_type type = 2;
     bytes payload = 3;
-    uint64 timestamp = 4;
+    uint64 tmst = 4; // internal timer of concentrator, only useful for reply timing
```
vihu: Can we rename this to `internal_conc_timer`? Or `conc_timer` Or something else perhaps a bit more descriptive?

src/packet.proto
Vagabond: it is typically in the range of 20-200ns but maybe?

src/packet.proto
Vagabond: ntp supplies them as floats

src/packet.proto
JayKickliter: It's sourced from UBX NAV-PVT which represents it as a u32

```tAcc ns Time accuracy estimate (UTC)```

File: src/packet.proto [hunk 4]
```diff
@@ -2,6 +2,23 @@ syntax = "proto3";
...
+message gps_info {
+    float horizontal_accuracy = 1; // meters this will take into account h3 precision and distance from asserted location
+    float vertical_accuracy = 2; // meters
+    int32 altitude = 3; // meters
+    uint64 time_accuracy = 4; //nanoseconds
+}
+
+// https://www.eecis.udel.edu/~mills/ntp/html/ntpq.html#system
+message ntp_info {
+    bool synced = 1;
```
vihu: So this means that it has synced with some ntp server? Don't see it in the ntp system link above, just wondering.

src/packet.proto
```diff
@@ -2,6 +2,23 @@ syntax = "proto3";
...
+    uint32 stratum = 2;
```
vihu: Maybe add a comment here that `stratum` is 1-15 as mentioned by ntp standard. Unsure if both are inclusive.

src/packet.proto
```diff
@@ -2,6 +2,23 @@ syntax = "proto3";
...
+    float sys_jitter = 3; // combined system jitter, in milliseconds
+    float clock_jitter = 4; // clock jitter, in milliseconds
```
vihu: nitpicking, rename to `clk_jitter` to follow the ntp standard

src/packet.proto
```diff
@@ -2,6 +2,23 @@ syntax = "proto3";
...
+    float clock_wander = 5; // clock frequency wander, in PPM
```
vihu: nitpicking, rename to `clk_wander` same as ntp

madninja:
@Vagabond are we leaving this open until progress is made on other timesources? or can this get closed?

PR #25, opened 2020-04-11
Add js integration

allenan:


madninja:
Awesome thanks @allenan! An chance you can get travis to pick it up as part of the build?

allenan:
@madninja added travis config for node js with a couple of tests

PR #29, opened 2020-05-07
State channel support

Vagabond:


src/blockchain_state_channel_v1.proto
```diff
@@ -42,7 +42,7 @@ message blockchain_state_channel_packet_v1 {
 
 message blockchain_state_channel_message_v1 {
     oneof msg {
-        blockchain_state_channel_packet_v1 packet = 1;
         blockchain_state_channel_response_v1 response = 2;
+        blockchain_state_channel_packet_v1 packet = 4;
```
madninja, line 46: Maybe leave a comment as to why this is more compatible? What compatibility do we need since this wasn't actually used anywhere yet?

Vagabond: It was used already, when I did the lorawan refactor I switched to using the state channel messages to ease this transition.

PR #31, opened 2020-05-11
Be more compatible with older message definitions

Vagabond:
For some reason we lost the compatibility conversions in the proto merge, this should restore them.

### DIFF
--- a/src/blockchain_state_channel_v1.proto
+++ b/src/blockchain_state_channel_v1.proto
@@ -31,7 +31,7 @@ message blockchain_state_channel_v1 {
 
 message blockchain_state_channel_response_v1 {
     bool accepted = 1;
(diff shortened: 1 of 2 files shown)


src/blockchain_state_channel_v1.proto
```diff
@@ -31,7 +31,7 @@ message blockchain_state_channel_v1 {
 
 message blockchain_state_channel_response_v1 {
     bool accepted = 1;
-    packet downlink = 2;
```
vihu, line 34: Just for my knowledge, the omission of other old fields (in numbering) is fine I'm assuming? This is just so that we don't break previously configured messges. Right?

Vagabond: yes

PR #32, opened 2020-05-12
Add price oracle protobuf definition

mrallen1:
This adds the price oracle message protobuf definition

### DIFF
--- a/src/blockchain_txn_price_oracle_v1.proto
+++ b/src/blockchain_txn_price_oracle_v1.proto
@@ -1,0 +1,9 @@
+syntax = "proto3";
+package helium;
+
+message blockchain_txn_price_oracle_v1 {
+    bytes public_key = 1;
+    uint64 price = 2;
+    uint64 block_height = 3;
+    bytes signature = 4;
+}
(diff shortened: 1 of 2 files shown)


src/blockchain_txn_price_oracle_v1.proto
```diff
@@ -0,0 +1,8 @@
+syntax = "proto3";
+package helium;
+
+message blockchain_txn_price_oracle_v1 {
+    bytes oracle_signature = 1;
```
Vagabond, line 5: one of these is probably the public key, and the other is the signature, right?

mrallen1: Yes that's right - the "oracle_signature" field is the signature _from_ the oracle validating its price estimate.

I thought the work flow would be something like:
1. Oracle estimates price
2. Oracle uses its private key to sign price
3. Price is expressed as cents
4. The price (as cents) goes in the price field
5. Signature of the price goes into "oracle_signature" field
6. The "signature" field here is supposed to be for sending the message into the blockchain?

src/blockchain_txn_price_oracle_v1.proto
```diff
@@ -0,0 +1,9 @@
+syntax = "proto3";
+package helium;
+
+message blockchain_txn_price_oracle_v1 {
+    bytes public_key = 1;
```
madninja, line 5: Just to verify.. no `fee` for oracle submissions?

PR #34, opened 2020-05-29
Add support for 868 channel mapping

andymck:
Proto PR for https://github.com/helium/miner/pull/405

### DIFF
--- a/src/blockchain_state_channel_v1.proto
+++ b/src/blockchain_state_channel_v1.proto
@@ -39,6 +39,18 @@ message blockchain_state_channel_packet_v1 {
     packet packet = 1;
     bytes hotspot = 2;
     bytes signature = 3;
+    enum Region {
+      US915 = 0;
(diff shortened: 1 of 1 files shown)


src/blockchain_state_channel_v1.proto
```diff
@@ -39,6 +39,18 @@ message blockchain_state_channel_packet_v1 {
     packet packet = 1;
     bytes hotspot = 2;
     bytes signature = 3;
+    enum Region {
```
vihu, line 42: I don't fully understand why `Region` enum is defined within the message stanza?

Vagabond: Why not?

vihu: Just curious is all, if it's fine here then let's get this in.

Vagabond: merge it

PR #35, opened 2020-06-04
Revert "Remove credits from sc proto"

Vagabond:
Reverts helium/proto#33

Apparently we have some records on-chain with credits in them.


Vagabond:
I've confirmed this fixes snapshot syncing, so will revert.

PR #36, opened 2020-06-05
Remove skewed proto

vihu:


evanmcc:
have we ever made one of these?  if so we need to keep it around no?

evanmcc:
github UI strikes again

vihu:
> have we ever made one of these? if so we need to keep it around no?

The only thing we've put on the chain is the `root_hash` and not the actual skewed tree

src/blockchain_state_channel_v1.proto
```diff
@@ -3,7 +3,7 @@ syntax = "proto3";
 package helium;
 
 import "packet.proto";
-import "skewed.proto";
+// import "skewed.proto";
```
madninja, line 6: Don't comment it out, just remove it

vihu: Updated in 83147e0

PR #38, opened 2020-06-11
Packet offer-purchase semantics

vihu:


src/blockchain_state_channel_v1.proto
```diff
@@ -32,29 +44,44 @@ message blockchain_state_channel_v1 {
...
-      IN865 = 8;
-    }
     Region region = 4;
 }
 
+message blockchain_state_channel_offer_v1 {
+    uint32 devaddr = 1;
+    bytes packet_hash = 2;
+    uint64 payload_size = 3;
+    uint32 seqnum = 4;
```
macpie, line 61: is this the `fcnt` (fcount) if so maybe name it fcount for consistency

src/blockchain_state_channel_v1.proto
```diff
@@ -32,29 +44,44 @@ message blockchain_state_channel_v1 {
...
+
+message blockchain_state_channel_purchase_v1 {
+    blockchain_state_channel_v1 sc = 1;
+    bytes hotspot = 2;
+    bytes signature = 3;
+    bytes packet_hash = 4;
+    Region region = 5;
+}
+
+message blockchain_state_channel_banner_v1 {
```
macpie, line 75: what is this used for?

src/blockchain_state_channel_v1.proto
vihu: @Vagabond ^

src/blockchain_state_channel_v1.proto
vihu: I will add a comment here, it's essentially used to communicate to the client hotspots whether the active state channel has altered or changed entirely

src/blockchain_state_channel_v1.proto
```diff
@@ -32,29 +44,44 @@ message blockchain_state_channel_v1 {
 message blockchain_state_channel_response_v1 {
     bool accepted = 1;
     packet downlink = 4;
+    bytes active_sc = 5;
```
Vagabond, line 47: Can't you nest the actual structure in here?

src/blockchain_state_channel_v1.proto
Vagabond: sure, why not

src/blockchain_txn_state_channel_close_v1.proto
```diff
@@ -9,4 +9,5 @@ message blockchain_txn_state_channel_close_v1 {
     bytes closer = 2;
     bytes signature = 3;
     uint64 fee = 4;
-}
\ No newline at end of file
+    blockchain_state_channel_v1 conflicts_with = 1;
```
vihu, line 12:
```suggestion
    blockchain_state_channel_v1 conflicts_with = 5;
```

src/blockchain_state_channel_v1.proto
```diff
@@ -32,29 +44,48 @@ message blockchain_state_channel_v1 {
 message blockchain_state_channel_response_v1 {
     bool accepted = 1;
     packet downlink = 4;
+    blockchain_state_channel_v1 active_sc = 5;
```
Vagabond, line 47: do we even need this field?

vihu: Hmm, I don't think we use it anywhere actually. I must have added it assuming we might. I'll remove it.

src/blockchain_state_channel_v1.proto
```diff
@@ -38,23 +50,41 @@ message blockchain_state_channel_packet_v1 {
...
+    bytes hotspot = 5;
+    bytes signature = 6;
+    Region region = 7;
+}
+
+message blockchain_state_channel_purchase_v1 {
+    blockchain_state_channel_v1 sc = 1;
+    bytes hotspot = 2;
+    bytes packet_hash = 3;
+    Region region = 4;
```
Vagabond, line 70: Why do we have region here?

PR #41, opened 2020-06-22
Add assert loc v2 proto

vihu:
Opening this to gather suggestions about what actually goes in this txn. Noteworthy detail is that `gateway_signature` has been excluded from here.

### DIFF
--- a/src/blockchain_txn.proto
+++ b/src/blockchain_txn.proto
@@ -37,6 +37,8 @@ import "blockchain_txn_transfer_validator_stake_v1.proto";
 import "blockchain_txn_unstake_validator_v1.proto";
 import "blockchain_txn_validator_heartbeat_v1.proto";
 
+import "blockchain_txn_assert_location_v2.proto";
+
 message blockchain_txn {
     oneof txn {
         blockchain_txn_add_gateway_v1 add_gateway = 1;
@@ -73,6 +75,7 @@ message blockchain_txn {
         blockchain_txn_validator_heartbeat_v1 val_heartbeat = 32;
         blockchain_txn_consensus_group_failure_v1 consensus_group_failure = 33;
         blockchain_txn_rewards_v2 rewards_v2 = 34;
+        blockchain_txn_assert_location_v2 assert_location_v2 = 35;
     }
 }
 
--- a/src/blockchain_txn_assert_location_v2.proto
+++ b/src/blockchain_txn_assert_location_v2.proto
@@ -1,0 +1,16 @@
+syntax = "proto3";
+package helium;
+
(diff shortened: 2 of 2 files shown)


ke6jjj:
I've  heard, informally, that there are some more items to add to the transaction. Can anyone remember what they are and post them here?

vihu:
> I've heard, informally, that there are some more items to add to the transaction. Can anyone remember what they are and post them here?

cc: @Vagabond

vihu:
@ke6jjj We were discussing about the extra items internally and it seems that it would be a good idea to add placeholders for antenna `tx/rx gain` and perhaps `elevation`, I believe all of these would be floats. It's okay if we don't use either of those from the get go but would be nice to have them here so we can update in the future without changing the protos. Thoughts?

Vagabond:
Yeah, I think at a minimum we want to add the gain levels.

Other things that might be interesting:

* Indoor/outdoor
* Radio model (eg SX1301 vs SX1302)

ke6jjj:
Happy to add the gain levels because that was the substance of a HIP I was going to propose! Elevation seems fine, too (as long as it is optional).

While I like the idea of the radio model because it stokes my personal interest in gathering performance data, I'm wondering how it will play out in the long term. Would it be an enumeration? How do we "onboard" new radios? Is it really something that the chain would eventually use for PoC?

As a counter argument, consider that since there is a centralized on-boarding server with a vendor "backchannel" for onboarding coupons, there should already be plenty of information about which radio model is in use on any particular hotspot, even if it isn't publicly available. If Helium needed the radio model purely for performance logging, surely it could be obtained that way, or a researcher like me could submit requests to Dewi to have access to the data in some form.

ke6jjj:
I'm going to advocate that if gain is added, it not be split into "tx_gain" versus "rx_gain". It should simply be "gain" and it should be understood that it applies to both tx and rx. We want to build an incentive for people to report their gain levels, but to get them to report accurately there must be a carrot and a stick.

The carrot is that a higher gain report will reduce your RSSI/SNR invalidations by providing an offset that will be applied to your RSSI reports. The stick, however, is that this benefit comes with the cost of having your TX power reduced accordingly. If we decouple these two then you can report a low TX gain and a high RX gain and completely defeat the purpose of the RSSI/SNR/FSPL curves entirely.

vihu:
> If we decouple these two then you can report a low TX gain and a high RX gain and completely defeat the purpose of the RSSI/SNR/FSPL curves entirely.

Good point, agreed. I'll update the proto to have one gain.

Vagabond (approved):

PR #46, opened 2020-07-22
Add MIC to offer msg

vihu:


vihu:
This is no longer required by router, closing.

PR #48, opened 2020-08-08
Add pending_close state to state channel messages

mrallen1:
Needed by https://github.com/helium/blockchain-core/pull/547


madninja:
@mrallen1 I'm going to assume this PR is no longer needed?

PR #51, opened 2020-09-15
Add serde [Rust]

lthiery:
This provides some serde attributes allowing the proto definitions to be leveraging when deserializing JSON from the API.

This should not affect current Rust proto bindings at all, but needs to be audited within that scope [as procedural macros are unhygenic](https://doc.rust-lang.org/reference/procedural-macros.html).

Of note is the deserializer for bs58 which tolerates null fields in the JSON. [A ticket against Blockchain API](https://app.clubhouse.io/hlm/story/8722/when-payer-is-null-populate-json-response-with-owner) has been filed which might allow for the deserializer to no longer tolerate null fields.


src/lib.rs
```diff
@@ -1,3 +1,196 @@
+mod origin {
```
JayKickliter, line 1: what's the reasoning behind having all these module blocks?

lthiery: good point. i think i started with a copy-paste example and just kept going. i should fix this

JayKickliter:
I see you just pushed to this PR. I take it that it's still alive?

lthiery:
> I see you just pushed to this PR. I take it that it's still alive?

Yea i still use this stuff in transaction analysis in Rust. I keep the branch alive for personal reasons, but I can close the PR and move this to a personal repo if @madninja prefers

uhrssophie (approved):

uhrssophie (approved):

madninja:
Can we close this @lthiery ?

lthiery:
> Can we close this @lthiery ?

Yessir, closed.

PR #53, opened 2020-09-24
Add rx1/2 window to helium packet

macpie:


madninja:
This can’t be right since it renumbers fields pretty drastically?

vihu:
@macpie Yeah, you can't change the proto like that without breaking backward compat. Instead just add `rx1_window` and `rx2_window` extra fields.

macpie:
This should be ok as it is only used for communication . Had a talk with @Vagabond about it

vihu:
> This should be ok as it is only used for communication . Had a talk with @Vagabond about it

Hmm, I guess that's fair, that packet(s) themselves never really show up in the ledger or any txn, so presumably it should be fine?

madninja:
That is fair.. but it would stop gateways that are not using the latest photo from being able to deliver packets right?

macpie:
Yes indeed this would need to be a coordinated update with router

macpie:
This is not to be merged until I got the rest done anyway (wanted to make this a draft but forgot).

Vagabond:
I would strongly recommend against renumbering fields in any case, it's not worth the extra explosion risk.

Vagabond (approved):

PR #57, opened 2020-12-13
compartmentalize the Rust

lthiery:
This pulls the Rust files from the `/src` directory into `helium-proto-rs`.


madninja:
I'm going to close this for now since compartmentalizing rust is nice but then it'll be the odd language out in the set of them

PR #62, opened 2021-02-24
No validator signatures

madninja:
Validator signatures are present to indicate key ownership, and as such only safety feature.

For simplicity of stake/unstake/transfer flows, remove these validator signature fields.

Also fixes up heartbeat block height to have a uint64 type


src/blockchain_txn_transfer_validator_stake_v1.proto
```diff
@@ -6,7 +6,6 @@ message blockchain_txn_transfer_validator_stake_v1 {
   bytes new_address = 2;
   bytes old_owner = 3;
   bytes new_owner = 4;
-  bytes new_validator_signature = 5;
   bytes old_owner_signature = 6;
```
evanmcc, line 9: reorder the rest?

PR #71, opened 2021-04-04
keep gateway service in one service

madninja:
Apologies for the whitespace changes.. I have proto auto-fmt turned on


andymck:
whilst not ideal on the erlang side, it will work fine.  

On the erlang side services map to modules, and then RPCs to functions within each module.  With routing and state channels now part of a single service the functional code for each will be intertwined.  Its not too bad with only 2 services but as we add more services it will start to get messy.

However if needs be and at that point I can add a functional "router" module which point each grouped RPCs ( ie state channels, routing ) to their own modules and maintain separation of concerns

andymck (approved):

PR #84, opened 2021-07-02
poc grpc requirements WIP

andymck:


src/blockchain_txn_poc_receipts_v2.proto
```diff
@@ -0,0 +1,14 @@
+syntax = "proto3";
+package helium;
+
+import "blockchain_poc_core_v1.proto";
+
+message blockchain_txn_poc_receipts_v2 {
```
Vagabond, line 6: If this is the same as receipts_v1 why are we duplicating it

jeffgrunewald: the data type for field 7 (request_block_hash vs. block_hash) is the same but will they be handled by the encoding/decoding code sufficiently differently as to justify a different versioned pb message?

jeffgrunewald: i believe we need to trash all open/active PoC txns that are mid-flight when the chain vars are switched over so we wouldn't have to worry about a txn being encoded by one code path and decoded by another so would it be sufficient to have the code generating/receiving these txns handle that last field appropriately based _solely_ on the chain vars? (i.e. look up a requesting block by hash vs. a challenge keys submission block by hash)

src/service/gateway.proto
```diff
@@ -17,6 +20,18 @@ enum close_state {
 
 /* general */
 
+message empty {}
+
+message gateway_success_resp {
+  bytes resp = 1;
```
madninja, line 26: can we make success_resp be  empty? including actual data seems to imply we're missing some actual types for the responses that have them. 
i.e. make 

```
message gateway_success_resp {}
```
and remove the `empty{}`

andymck: That is on my to-do list but was leaving it til later..IE before the rust client comes along....what is there now works fine for Erlang client

madninja: yeah just that once validators are deployed with this we'll have issues removing fields there

madninja: Especially since we're already adjusting the protos again?

PR #92, opened 2021-09-14
add additional fields to is_active resp

andymck:
Adds fields expiry_at_block and original_dc_amount

Linked PRs:
https://github.com/helium/sibyl/pull/13

### DIFF
--- a/src/service/gateway.proto
+++ b/src/service/gateway.proto
@@ -36,6 +36,8 @@ message gateway_sc_is_active_resp_v1 {
   bytes sc_id = 1;
   bytes sc_owner = 2;
   bool active = 3;
+  uint64 sc_expiry_at_block = 4;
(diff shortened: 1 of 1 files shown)


src/service/gateway.proto
```diff
@@ -36,6 +36,8 @@ message gateway_sc_is_active_resp_v1 {
   bytes sc_id = 1;
   bytes sc_owner = 2;
   bool active = 3;
+  uint64 sc_expiry_at_block = 4;
```
madninja, line 39: why prefix with `sc_` here if `active is not?

andymck: the way i have it is any field derived from the SC itself is prefixed with the sc, and active in this case is not taken from the SC but signals the SC does exist

madninja: Got it.. good explanation, thanks

PR #98, opened 2021-10-12
Adds state channel diffs to purchases

madninja:
So purchase can either have a full state channel or a state channel diff which is a set of differences to a given state channel identified by its id/nonce.

Note that diffs are only ever additive to ensure causality when a diff is presented


src/blockchain_state_channel_v1.proto
```diff
@@ -72,6 +72,38 @@ message blockchain_state_channel_purchase_v1 {
     bytes hotspot = 2;
     bytes packet_hash = 3;
     Region region = 4;
+    blockchain_state_channel_diff sc_diff = 5;
+}
+
+message blockchain_state_channel_diff {
+    bytes id = 1;
+    uint64 nonce = 2;
```
Vagabond, line 80: what is this nonce for?

src/blockchain_state_channel_v1.proto
```diff
@@ -72,6 +72,38 @@ message blockchain_state_channel_purchase_v1 {
...
+    bytes signature = 3;
+    repeated blockchain_state_channel_diff_entry = 4;
+}
+
+enum blockchain_state_channel_diff_entry {
+    oneof entry {
+        blockchain_state_channel_diff_update_nonce nonce = 1;
```
Vagabond, line 87: hoist this as a uint64 directly?

madninja: Yeah that was what the nonce above was for

src/blockchain_state_channel_v1.proto
Vagabond: I would frame this as a nonce *increment*, so it can only go up.

src/blockchain_state_channel_v1.proto
```diff
@@ -65,15 +65,44 @@ message blockchain_state_channel_offer_v1 {
...
+message blockchain_state_channel_diff_v1 {
+    bytes id = 1;
+    uint64 add_nonce = 2;
+    bytes signature = 3;
+    repeated blockchain_state_channel_diff_entry_v1 diffs = 4;
+}
+
+message blockchain_state_channel_diff_entry_v1 {
+    oneof entry {
+        blockchain_state_channel_diff_append_summary_v1 append = 2;
```
madninja, line 88: Maybe renumber these 1 and 2?

PR #100, opened 2021-10-14
Remove rustfmt in tonic-build dependency by default to allow production builds without rustfmt

mullerch:
Actually, by default, rustfmt is needed for all builds. As rustfmt is here only to prettify the generated proto files (very useful for debug), it can be removed for production builds.

I could not find a way to differentiate the build-dependency depending on the profile (dev, release).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,5 +17,5 @@ serde_derive = "1"
 serde_json = "1"
 
 [build-dependencies]
-tonic-build = {version = "0", optional = true, default-features = false, features = ["transport"] }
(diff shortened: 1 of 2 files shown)


mullerch:
What about now?

madninja:
> What about now?

nice, does that work for you?

mullerch:
Yes that works fine. Looks like unused functions do not imply a dependency.

madninja:
> Yes that works fine. Looks like unused functions do not imply a dependency.

Yeah I believe it's literally shelling out to rustfmt if format is set to true

madninja:
Thanks @mullerch !

PR #101, opened 2021-10-19
Rust: pub use prost::Encode/DecodeError

lthiery:
prost::Message provides functions which return `Result<_, Encode/DecodeError>`. It's easy to derive application level errors from these errors if we make these types accessible to the client. Otherwise, the client needs to import the right version of `prost` themselves.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,6 +1,8 @@
 include!(concat!(env!("OUT_DIR"), "/helium.rs"));
 pub use blockchain_txn::Txn;
 pub use prost::Message;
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1,6 +1,8 @@
 include!(concat!(env!("OUT_DIR"), "/helium.rs"));
 pub use blockchain_txn::Txn;
 pub use prost::Message;
+pub use prost::{DecodeError, EncodeError};
```
madninja, line 4: use the combined pub use Prost::{} approach please..

PR #116, opened 2022-01-25
grpc txn submit protos reqs

andymck:
Provides proto side of https://github.com/helium/blockchain-core/pull/1216


src/service/gateway.proto
```diff
@@ -36,6 +86,10 @@ message gateway_resp_v1 {
     gateway_config_resp_v1 config_resp = 8;
     gateway_config_update_streamed_resp_v1 config_update_streamed_resp = 9;
     gateway_validators_resp_v1 validators_resp = 12;
+    gateway_success_resp success_resp = 13;
```
madninja, line 89: Neither success_resp of error_resp are needed in this PR right?

madninja (commented):
Run the formatter, and remove some unneeded resp messages?

src/service/gateway.proto
```diff
@@ -19,6 +20,55 @@ enum close_state {
...
+message gateway_submit_txn_resp_v1 { bytes key = 1;
+  routing_address validator = 2;
+}
+
+message gateway_query_txn_req_v1 { bytes key = 1; }
+message gateway_query_txn_resp_v1 {
+  txn_status status = 1;
+  repeated acceptor acceptors = 2;
+  repeated rejector rejectors = 3;
+  bytes details = 4;
```
madninja, line 68: what would go in here? Like a failed_reason?

madninja (commented):
just one question.. and run the formatter if you could

src/service/transaction.proto
```diff
@@ -0,0 +1,47 @@
+syntax = "proto3";
+
+import "blockchain_routing_address.proto";
+import "blockchain_txn.proto";
+
+enum txn_status {
+  pending = 0;
+  cleared = 1;
```
andymck, line 8: We wont need cleared here anymore as the txn mgr wont be hold this in the revised version of the backend changes

src/service/transaction.proto
```diff
@@ -0,0 +1,48 @@
...
+  uint64 height = 1;
+  bytes reason = 2;
+  bytes pub_key = 3;
+}
+
+message txn_submit_req_v1 { blockchain_txn txn = 1; }
+
+message txn_submit_resp_v1 {
+  bytes key = 1;
+  uint64 recv_height = 2;
```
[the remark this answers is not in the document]
jeffgrunewald:
```suggestion
  optional uint64 recv_height = 2;
```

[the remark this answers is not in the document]
andymck: > 

We could alternatively default the height to zero in such cases ? 

Do optional fields result in the field being dropped if not set or having a value undefined ?

src/service/transaction.proto
```diff
@@ -0,0 +1,48 @@
+syntax = "proto3";
```
andymck, line 1: for some baffling reason this proto fails to compile

===> Writing /Users/andymck/helium/blockchain-core/src/grpc/autogen/server/transaction_pb.erl
===> Error compiling proto file transaction.proto in msg txn_query_req_v1, field key: undefined reference  blockchain_txn

src/service/transaction.proto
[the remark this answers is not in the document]
jeffgrunewald: proto3 doesn't support anything like a null value but it does allow you to omit the field altogether if you mark it as optional and then it's up to the language implementation to deal with that the best way it knows how. i the case of gpb it turns those into `undefined` so i _think_ that makes the most sense here

madninja (approved):

PR #117, opened 2022-02-09
Region Params requirements

andymck:
Adds proto def for region params RPC

### DIFF
--- a/src/service/gateway.proto
+++ b/src/service/gateway.proto
@@ -14,21 +14,7 @@ enum close_state {
   close_state_dispute = 3;
 }
 
-enum error_code {
-  generic = 0;
-  invalid_signature = 1;
-  not_asserted = 2;
-  unknown_region = 3;
-}
-
 /* general */
-
-message gateway_success_resp {}
-
-message gateway_error_resp {
-  error_code error = 1;
-  bytes details = 2;
-}
 
 message gateway_config_req_v1 { repeated string keys = 1; }
 message gateway_config_resp_v1 { repeated blockchain_var_v1 result = 1; }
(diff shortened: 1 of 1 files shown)


src/service/gateway.proto
```diff
@@ -13,8 +14,25 @@ enum close_state {
   close_state_dispute = 3;
 }
 
+enum error_code {
+  error_generic = 0;
```
madninja, line 18: why prefix it all with `error_` if it's already in the error_code enum?

src/service/gateway.proto
```diff
@@ -13,8 +14,25 @@ enum close_state {
...
+  error_invalid_signature = 1;
+  error_not_asserted = 2;
+  error_unknown_region = 3;
+}
+
 /* general */
 
+message gateway_success_resp {
```
madninja, line 26: not needed as part of this PR right? If so, what are they actually useful for? Generic "bytes" on success is quite vague?

src/service/gateway.proto
```diff
@@ -13,8 +14,25 @@ enum close_state {
...
+  bytes resp = 1;
+  bytes details = 2;
+}
+
+message gateway_error_resp {
```
madninja, line 31: ditto? Not needed in this PR?

src/service/gateway.proto
andymck: no hard reason other than the error prefix makes it easier to convey the msgs are a control set on the client side

src/service/gateway.proto
andymck: success resp is not required but the error resp is required and these two go together.

src/service/gateway.proto
andymck: see above

src/service/gateway.proto
andymck: success_resp is used for RPCs where there is no defined response payload to be returned.  For example when submitting a POC witness or receipt report.  In that example 'resp' has a value of ok....prob does make sense to do something else there than bytes

madninja: how about just an empty success response?

andymck: yeah prob as much as it needs.  will roll with that

PR #119, opened 2022-02-13
Python

thecardcheat:
Commit message has details of protobuf changes. I am unsure of the consequence for existing clients but the name change to avoid the namespace conflict seems breaking.


.github/workflows/python.yml
```diff
@@ -0,0 +1,34 @@
...
+          PROTOC_ZIP: "protoc-3.19.4-linux-x86_64.zip"
+        run: wget https://github.com/protocolbuffers/protobuf/releases/download/v$PROTOC_VERSION/$PROTOC_ZIP && sudo unzip -o $PROTOC_ZIP -d /usr/local bin/protoc && sudo unzip -o $PROTOC_ZIP -d /usr/local 'include/*' && sudo chmod a+x /usr/local/bin/protoc && rm -f $PROTOC_ZIP
+      - name: Install requirements
+        run: pip install -r requirements.txt
+      - name: Build python module
+        run: mkdir generated && protoc -I src --python_betterproto_out=generated/ src/*.proto src/service/*.proto
+      # Appears related to https://github.com/danielgtaylor/python-betterproto/issues/238 but we need the latest beta of
+      # betterproto to build src/service/*.proto
+      - name: Trim python module
+        run: sed -i '/from . import /d' generated/helium/__init__.py
```
madninja, line 29: Missing a quote here?

src/blockchain_txn_poc_receipts_v1.proto
```diff
@@ -2,8 +2,8 @@ syntax = "proto3";
 package helium;
 
 enum origin {
-  p2p = 0;
-  radio = 1;
```
madninja, line 6: Please don’t rename fields

src/service/gateway.proto
```diff
@@ -2,7 +2,6 @@ syntax = "proto3";
 
 package helium;
 
-import "blockchain_state_channel_v1.proto";
```
madninja, line 5: Why are you removing this?

src/blockchain_txn_poc_receipts_v1.proto
thecardcheat: This was done due to the following error:

> radio.proto:3:9: "helium.radio" is already defined (as something other than a package) in file "blockchain_txn_poc_receipts_v1.proto".

This appears to be an issue with protobuf namespacing, I'm not sure why things that aren't `protoc` didn't care and I made this change based on a recommendation in an SO answer from one of protobuf's original authors about how to handle these conflicts. If there is an alternate method that wouldn't break the naming that'd be preferable, but I am unaware of what that would be.

src/service/gateway.proto
thecardcheat: `protoc` raised a warning that this import was unused.

.github/workflows/python.yml
thecardcheat: I don't believe so, the artifact had the lines removed as intended when I downloaded it to validate.

.github/workflows/python.yml
```diff
@@ -0,0 +1,34 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python 3.10
+        uses: actions/setup-python@v2
+        with:
+          python-version: "3.10"
+      - name: Install protoc
```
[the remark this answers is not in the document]
thecardcheat: Changed to apt.

src/blockchain_txn_poc_receipts_v1.proto
madninja: The problem is that it'll break existing erlang and rust users of this message.

thecardcheat: Will try to dig into those a little bit more to compare, thanks for the tip.

madninja:
@thecardcheat Are you still interested in this PR? if so we need to get that comment addressed. If not, can this PER get closed?

ccall48:
I would be very interested in a working python version, as compiling the project threw the same errors on me on trying to run it (also a given i am very green with grpc atm). However with Openlns now a thing it will let me integrate backend changes to route ids directly through python instead of another language i do not know.

PR #123, opened 2022-03-11
include attestation data in receipt and witness reports

andymck:
proto side of https://github.com/helium/miner/pull/1481

### DIFF
--- a/src/blockchain_poc_core_v1.proto
+++ b/src/blockchain_poc_core_v1.proto
@@ -6,6 +6,14 @@ enum origin {
   radio = 1;
 }
 
+message attestation {
+  uint64 height = 1;
+  uint64 block_time = 2;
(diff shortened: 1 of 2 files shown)


src/blockchain_poc_core_v1.proto
```diff
@@ -4,6 +4,15 @@ package helium;
 enum origin {
   p2p = 0;
   radio = 1;
+  grpc = 2;
+}
+
+message attestation {
+  uint64 height = 1;
+  uint64 block_time = 2;
+  uint64 block_age = 3;
```
madninja, line 13: how is block_age being used here?

madninja:
can we close this @andymck ?

andymck:
yeah unlikely this is going to go anywhere now

PR #124, opened 2022-03-14
Andymck/poc grpc plus vals as chall

andymck:
Proto side of https://github.com/helium/miner/pull/1469

### DIFF
--- a/src/blockchain_block_v1.proto
+++ b/src/blockchain_block_v1.proto
@@ -13,7 +13,7 @@ message blockchain_seen_vote_v1 {
 }
 
 message blockchain_poc_key {
-  uint32 pos = 1;
+  bytes address = 1;
   bytes key = 2;
 }
 


src/service/gateway.proto
```diff
@@ -17,6 +20,18 @@ enum close_state {
 
 /* general */
 
+message empty {}
+
+message gateway_success_resp {
+  bytes resp = 1;
```
madninja, line 26: remember we talked about not having details in the success resp since if there are details it should really be a specific resp message. And if it's empty then what's the difference between empty and success_resp

src/service/gateway.proto
```diff
@@ -17,6 +20,18 @@ enum close_state {
...
+  bytes details = 2;
+}
+
+message gateway_error_resp {
+  bytes error = 1;
```
madninja, line 31: would it be possible to have the error be an actual enum here? it'd document and be more specific about what the possible errors are.

src/service/gateway.proto
andymck: yeah i did actually change these but it must be on another branch.  Will fix

src/service/gateway.proto
andymck: yeah i did actually change these but it must be on another branch.  Will fix

PR #125, opened 2022-03-15
add final flag to sync blocks message

joecaswell:
Add flag to blockchain_sync record to indicate final batch of blocks


madninja:
Huh, could you explain how this is useful?

madninja (approved):

PR #126, opened 2022-03-17
Pre-generate erlang grpc client modules alongside their proto files

jeffgrunewald:
By pre-generating the (erlang) client modules generated from these proto files, we can more easily review and inspect the generated code and save build times on downstream applications by not compiling them on the fly with every new build, especially when their definition will remain relatively stable most of the time.

This initial change introduces a minimal directory structure for two applications that can immediately consume this change in the format that is most easily added to a rebar.config dependencies block. They are structured by consuming application to allow fine-grained control over which proto services are consumed by which application. We should review automation around how to routinely compile and build these modules automatically alongside changes to the proto files.


andymck:
@jeffgrunewald is this able to accommodate a top level application such as miner having multiple grpc config files ?  For example the POC miner branch has two such config files, each specifying different config options and generating different sets of PB files.

The options in the config do ensure the generated PB files have unique filenames

jeffgrunewald:
sadly @andymck when actually _running_ the application the pre-generated client seems to be causing problems so this will need to wait until we pass the new release milestone to dig into and figure out what's wrong but looking at the config as it stands this _should_ work for all proto client definitions in a given top-level application so long as they all share the same `gpb` options. if they don't (and i believe your current miner branch configs have some differences from those the local api is using) then this would need to be structured differently to produce the result @evanmcc is looking for of a per-application dependency profile. once we've got time for that and he's available again we should look into this further and see if it's possible and worth pursuing

PR #127, opened 2022-03-22
Extend local rpc api with add_gateway_txn

jeffgrunewald:
The current hotspots allow for (ebus) rpc communication with the miner to allow generating and returning a signed `blockchain_txn_add_gateway_v1` transaction so the rust light gateway will need to support this functionality as well. This rpc provides a bridge api between the legacy call (which accepts an owner, payer, amount, and fee) and the newer api currently defined in the light gateway where staking fees are immediately known based on the mode of the device and do not need to be calculated from the chain.

### DIFF
--- a/src/service/local.proto
+++ b/src/service/local.proto
@@ -38,8 +38,11 @@ message region_res { int32 region = 1; }
 message add_gateway_req {
   string owner = 1;
   string payer = 2;
-  uint64 amount = 3;
-  uint64 fee = 4;
+  enum staking_mode {
+    dataonly = 0;
+    full = 1;
+    light = 2;
+  }
 }
 
 message add_gateway_res { bytes add_gateway_txn = 1; }
--- a/src/service/local.proto
+++ b/src/service/local.proto
@@ -35,14 +35,15 @@ message height_res {
 message region_req {}
 message region_res { int32 region = 1; }
 
+enum staking_mode {
+  dataonly = 0;
+  full = 1;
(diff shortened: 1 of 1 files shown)


src/service/local.proto
```diff
@@ -35,11 +35,24 @@ message height_res {
 message region_req {}
 message region_res { int32 region = 1; }
 
+message add_gateway_req {
+  string owner = 1;
```
madninja, line 39: these should be bytes to represent the pubkey_bin, not the b58 version (which I think you're implying here)

madninja (commented):
Once nit, otherwise looks good

src/service/local.proto
jeffgrunewald: should they be the b58 pubkeys? i see now that's what we've been sending over the ebus wire

madninja: that may be true for ebus, but in grpc we send the pubkey_bin around since it has a binary type

src/service/local.proto
```diff
@@ -35,14 +35,15 @@ message height_res {
 message region_req {}
 message region_res { int32 region = 1; }
 
+enum staking_mode {
```
madninja, line 38: Maybe overly cautious, but rename to `gateway_staking_mode` since staking is a common concept

madninja (changes requested):

madninja (approved):

PR #132, opened 2022-05-13
update region enum

mikev:
AS923_1B defines a Region supporting Malaysia.  It is illegal in Malaysia to use any frequency above 924 Mhz, hence the reason for this new Region.  The AS923_1B defines Uplink frequencies at [922.0, 922.2, 922.4, 922.6, 922.8, 923.0, 923.2, 923.4]

CD900_1A defines a new testnet Region.  This will be used for testing new Region parameters.  It is centered in the ocean at the Challenger Deep which is the lowest sea floor depth on Earth - https://en.wikipedia.org/wiki/Challenger_Deep

CN779 is officially deprecated by the LoRa Alliance and should never be used.

Update to gateway-rs
https://github.com/helium/gateway-rs/pull/250


src/region.proto
```diff
@@ -7,6 +7,7 @@ enum region {
   EU868 = 1;
   EU433 = 2;
   CN470 = 3;
+  // CN779 is deprecated.  Please remove.
```
madninja, line 10: leave the deprecated mark in there but you can't just remove the entry

mikev: In theory removing enum entries is supported.  As long as the physical number is not re-used.
https://developers.google.com/protocol-buffers/docs/proto

vihu: Its not only about whether its supported by protobuf but about whether it gets used in some critical path in core/miner (this is also true for any other field in other protos btw). That said I'm pretty sure CN779 isn't used anywhere but having it here also doesn't hurt.

If at some point we are to change the region lookup mechanism we can be more thorough.

madninja: In addition changing existing names will cause other tools to break (like gateway-rs)

mikev: OK.  I removed the 'remove' suggestion.

madninja (approved):

oberchoo:
Yes, in Malaysia, the allowed ISM band for 923 is 919 - 923MHz (a band) and 923 - 924MHz (another band). It will be great to see this AS923-1B that shifts the operating freq slightly lower. For more information please refer here: https://www.mcmc.gov.my/skmmgovmy/media/General/pdf/Class-Assignment-No-1-of-2021_.pdf (page 10 of 74)

PR #133, opened 2022-05-18
HIP XX Experiments

vihu:
TBD...


src/blockchain_txn_payment_v2.proto
```diff
@@ -1,12 +1,15 @@
...
 package helium;
 
+import "blockchain_token_type_v1.proto";
+
 message blockchain_txn_payment_v2 {
   bytes payer = 1;
   repeated payment payments = 2;
   uint64 fee = 3;
   uint64 nonce = 4;
   bytes signature = 5;
```
jeffgrunewald, line 11: So we think payment txns will always have to be one or more non-zero payments of the same token type, rather than adding a token_type to the payment record and being able to mix and match?

src/blockchain_ledger_entry_v2.proto
```diff
@@ -0,0 +1,15 @@
+syntax = "proto3";
+package helium;
+
+message blockchain_ledger_entry_v2 {
+  // NOTE: might need to add owner:binary here for portability reasons,
+  // leaving that out for now to revisit later.
+  uint64 hnt_nonce = 1;
```
jeffgrunewald, line 7: Have these subtypes been stable across hst and hnt or should these be broken into subrecords, like:
```
ledger_entry_v2{
  hnt_entry hnt_entry = 1,
  hst_entry hst_entry = 2,
  …
}

hnt_entry{
  uint64 balance = 1,
  uint64 nonce = 2
}

hst_entry{
  uint64 balance = 1,
  uint64 nonce = 2
}
…
```

src/blockchain_txn_payment_v2.proto
vihu: yeah, I don't think we want to mix and match in the individual payment record. Probably best to just have top level token_type specified in the payment txn itself. Unless there's some really compelling reason to do so. Mostly just saying to keep things clean-ish txn level

src/blockchain_ledger_entry_v2.proto
vihu: Sure, I'll break these out. But we probably would keep these static, not sure how well a generic version translates to rocks storage. I doubt these would be updated _very_ frequently :slightly_smiling_face:

src/blockchain_txn_payment_v2.proto
jeffgrunewald: The one edge-case I can think of would be balance-clearing txns. If the intent for those is to just clear a single token-type that should work but if the intent is to “burn out my account” then we’d need to generate an enumerable higher-level operation to create a clearing txn for each held token type

abhay: I think the token type should be on the individual payment and you should be able to transfer multiple.

madninja: I agree with @abhay here

src/blockchain_ledger_entry_v2.proto
madninja: I thought you were convinced it'd be better to have the static ledger fields instead of the generic "list of ledger entries" which adds ore complexity on update/insert

vihu: > I thought you were convinced it'd be better to have the static ledger fields instead of the generic "list of ledger entries" which adds ore complexity on update/insert

Oh, I wanted to avoid having `blockchain_ledger_entries_v2`, this one would work relatively okay. By static I meant not having generic `token_type` in the `ledger_entry` record.

src/blockchain_txn_payment_v2.proto
vihu: Hmm, alright then. I'll tweak the proto to reflect that individual payment has the token type instead.

src/blockchain_txn_add_network_v1.proto
```diff
@@ -0,0 +1,17 @@
...
+    blockchain_token_type_v1 token_type = 1;
+    
+    bytes subnetwork_key = 2;
+    repeated bytes reward_server_keys = 3;
+    uint64 premine = 4;
+
+    bytes network_signature = 5;
+    bytes subnetwork_signature = 6;
+
+    uint64 fee = 7;
```
madninja, line 16: Why have a fee here? A new network can only be added by the master key(s) right? Would not require fees?

src/blockchain_ledger_entry_v2.proto
```diff
@@ -0,0 +1,16 @@
+syntax = "proto3";
+package helium;
+
+import "blockchain_token_type_v1.proto";
```
jeffgrunewald, line 4: is this import needed?

src/blockchain_txn_token_convert_v1.proto
```diff
@@ -0,0 +1,15 @@
...
+package helium;
+
+import "blockchain_token_type_v1.proto";
+
+message blockchain_txn_token_convert_v1 {
+    blockchain_token_type_v1 token_type = 1;
+    bytes account = 2;
+    uint64 amount = 3;
+
+    bytes signature = 4;
```
vihu, line 11: Rename to `account_signature` maybe?

src/blockchain_txn_add_subnetwork_v1.proto
```diff
@@ -0,0 +1,15 @@
+syntax = "proto3";
+package helium;
+
+import "blockchain_token_type_v1.proto";
+
+message blockchain_txn_add_subnetwork_v1 {
+    blockchain_token_type_v1 token_type = 1;
```
jeffgrunewald, line 7: should the token types supported by the L1 helium network, including HNT and HST be separately tracked from subnetwork token types? This implies that HNT and HST could be valid token types for a subnetwork but that doesn't seem correct.

what i'm suggesting is perhaps two different enumerables, one that includes _all_ token types for purposes like payments and conversions and another for things like subnets where it seems like the top-level L1 tokens aren't in play?

src/blockchain_txn_token_convert_v1.proto
```diff
@@ -0,0 +1,15 @@
+syntax = "proto3";
+package helium;
+
+import "blockchain_token_type_v1.proto";
+
+message blockchain_txn_token_convert_v1 {
+    blockchain_token_type_v1 token_type = 1;
```
jeffgrunewald, line 7: is this always a one-way conversion, like HNT -> X? else how does it indicate what token is being converted into what other token?

src/blockchain_ledger_entry_v2.proto
vihu: Fixed 7328993

src/blockchain_txn_add_subnetwork_v1.proto
vihu: Yeah, that's a valid point. However, we can invalidate at txn validity level as well (which is what we're doing right now in the core branch) so not _really_ a huge deal. Also, given the time constraints I don't feel the need to change right away.

src/blockchain_txn_token_convert_v1.proto
```diff
@@ -0,0 +1,15 @@
+syntax = "proto3";
+package helium;
+
+import "blockchain_token_type_v1.proto";
+
+message blockchain_txn_token_convert_v1 {
```
abhay, line 6:
```suggestion
message blockchain_txn_token_redeem_v1 {
```

Wondering if `redeem` is better than `convert` here since the action is to redeem a MOBILE (for example) for the appropriate amount of HNT in the treasury.

src/blockchain_txn_add_subnetwork_v1.proto
```diff
@@ -0,0 +1,15 @@
+syntax = "proto3";
+package helium;
+
+import "blockchain_token_type_v1.proto";
+
+message blockchain_txn_add_subnetwork_v1 {
+  blockchain_token_type_v1 token_type = 1;
+
+  bytes subnetwork_key = 2;
+  repeated bytes reward_server_keys = 3;
```
abhay, line 10: should this be a list of keys or a 1/N multisig (where N can start as 1 for now) but can be changed later if we want to add / migrate keys.

evanmcc: no, a list is more appropriate until we change the model to one where we have multiple servers double checking each other, which is not what we have right now.

src/blockchain_txn_token_convert_v1.proto
evanmcc: we don't implement this txn in the work so far, so we can change it here or later, I don't have an opinion.

evanmcc (approved):

src/blockchain_txn_token_convert_v1.proto
vihu: Addressed in [49b084f](https://github.com/helium/proto/pull/133/commits/49b084f93a223bac7801f1be606c06d2e9ad2b74)

Ljzn:
Hi, I'm working on integrating helium into our wallet. I noticed a new field `token_type` be added to the "payment_v2" transaction. This could cause a breaking change to the user of Helium REST API, for example: with the old API, we assume that all `payment_v2` transaction is moving HNT. If we didn't be notified that there is a new "token_type" filed, we may handle the token payment incorrectly.

I suggest a new tx type such as `payment_v3` would fit this. Or do you have some advice for REST API users to prevent such conditions?

vihu:
> Hi, @vihu I'm working on integrating helium into our wallet. I noticed a new field `token_type` be added to the "payment_v2" transaction. This could cause a breaking change to the user of Helium REST API, for example: with the old API, we assume that all `payment_v2` transaction is moving HNT. If we didn't be notified that there is a new "token_type" filed, we may handle the token payment incorrectly.
> 
> I suggest a new tx type such as `payment_v3` would fit this. Or do you have some advice for REST API users to prevent such conditions?

I am pretty sure that the new field `token_type` is optional and has a default value of `hnt` so it should be pretty safe. I would suggest just getting in the new protobuf definitions and updating the necessary APIs. Typically for version bumps, the change has to be substantial (in terms of validation and absorption of the transaction on the blockchain).

Ljzn:
> I am pretty sure that the new field token_type is optional and has a default value of hnt so it should be pretty safe. I would suggest just getting in the new protobuf definitions and updating the necessary APIs. Typically for version bumps, the change has to be substantial (in terms of validation and absorption of the transaction on the blockchain).

I understand that this protocol modification is reasonable and has no impact on constructing a transaction. But many exchanges and wallets use APIs to download blocks, then filter out related transactions, and use JSON instead of protobuf (eg https://docs.helium.com/api ). So when an optional "token_type" is added to the protocol, the code doesn't throw any errors (unless we check the number of keys in the JSON return value, which people don't usually do), but the nature of the transaction changes (We thought it was an HNT transaction, but it may actually be other tokens)

PR #134, opened 2022-05-24
add initial subnetwork txns

evanmcc:


src/blockchain_txn_add_network_v1.proto
```diff
@@ -0,0 +1,14 @@
+syntax = "proto3";
+
+package helium;
+
+message blockchain_txn_add_network_v1 {
+    bytes treasury_key = 1;
+    repeated bytes rollup_keys = 2;
+    uint64 premine = 3;
```
vihu, line 8: may need token_type for premine

PR #135, opened 2022-05-25
Add version rpc to gateway service

madninja:


jeffgrunewald (approved):
Any reason not to just include `version` as a standard field on the basic `gateway_resp_v1` payload?

andymck (approved):

PR #141, opened 2022-06-06
Define follower service and req/resp

jeffgrunewald:
Required proto definition for https://github.com/helium/blockchain-node/pull/139. Defines the proto service for the blockchain-node follower to stream transaction information to consumers for off-chain activities.

### DIFF
--- a/src/service/follower.proto
+++ b/src/service/follower.proto
@@ -10,24 +10,13 @@ message follower_txn_stream_req_v1 {
   repeated string txn_types = 3;
 }
 
-message follower_txn_streamed_resp_v1 {
+message follower_txn_stream_resp_v1 {
   uint64 height = 1;
-  uint64 epoch = 2;
-  bytes txn_hash = 3;
-  blockchain_txn txn = 4;
-}
-
-message follower_resp_v1 {
-  uint64 height = 1;
-  bytes signature = 2;
-  uint64 block_time = 3;
-  uint64 block_age = 4;
-  oneof event {
-    follower_txn_streamed_resp_v1 txn_event = 5;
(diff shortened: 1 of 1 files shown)


src/service/follower.proto
```diff
@@ -0,0 +1,33 @@
...
+  blockchain_txn txn = 4;
+}
+
+message follower_resp_v1 {
+  uint64 height = 1;
+  bytes signature = 2;
+  uint64 block_time = 3;
+  uint64 block_age = 4;
+  oneof event {
+    follower_txn_streamed_resp_v1 txn_event = 5;
```
madninja, line 26: why have this enum?

jeffgrunewald: I can pull it out but I didn’t know if this was the only instance of a follower we might end up needing in future

andymck (approved):

madninja:
> > and check your formatting.. it's failing lint
> 
> I saw that; I see what lines it’s complaining about but I can’t figure out what it doesn’t like about them. Is this default format style documented somewhere (action says it should default to llvm)

just clang-format.. I use the vscode plugin and it "just works" :-)

src/service/follower.proto
madninja: > I can pull it out but I didn’t know if this was the only instance of a follower we might end up needing in future

if it is we'd add another rpc for it or change it later.. No need to over design here

madninja:
Actually, I'm unsure we need to do al the same things a remote rpc service would.. why sign and include the height/age in every streamed txn. The idea for this follower was that it'd be _local_ and trusted

PR #151, opened 2022-07-07
Add rewards_v3 proto

vihu:
Summary
----
This is pretty much the exact same thing as `rewards_v2`, I'm adding this for consistency reasons (since we'll be doing `rewards_v3` transaction separately in blockchain-core. Let me know if this is not required and we want to use the proto for `rewards_v2` for `blockchain_txn_rewards_v3`.

Also please let me know if we should add more stuff here for future proofing.


madninja:
can we close this @vihu ?

vihu:
Closing as this is deprecated due to design changes.

PR #154, opened 2022-08-02
initial pass at offchain lora protos

andymck:


src/service/poc_lora.proto
```diff
@@ -0,0 +1,64 @@
...
+}
+
+message poc_lora_valid {
+  bytes key = 1;
+  poc_lora_path_element path = 2;
+}
+
+message poc_lora_reject {
+    bytes key = 1;
+    bytes reject_reason = 2;
```
madninja, line 47: I'd like reject reason to be an enum of possible error instead of random bytes?

src/service/poc_lora.proto
```diff
@@ -0,0 +1,64 @@
...
+    bytes reject_details = 3;    
+    oneof msg {
+        poc_lora_beacon_req_v1 beacon = 4;
+        poc_lora_witness_req_v1 witness = 5;
+  }
+}
+
+message poc_lora_path_element {
```
madninja, line 55: why is this a "path element". Is there a better name for this?

src/service/poc_lora.proto
```diff
@@ -0,0 +1,64 @@
+syntax = "proto3";
+
+package helium.poc_lora;
+
+message poc_lora_beacon_req_v1 {
+  bytes poc_key = 1;
```
madninja, line 6: Explain what a poc key is here?

andymck: This would be similar to the old onionkeyhash.  I am assuming a beacon will still include some sort of ID which a witness can derive during its decrypt of the broadcast and include in its report.  Perhaps the packet_hash is sufficient on its own to serve as an ID but a specific ID could still be useful to for tracing purposes

src/service/poc_lora.proto
andymck: dejavu, but yeah will add that

src/service/poc_lora.proto
andymck: used old naming convention, will come up with something better suited to the new order

src/service/poc_lora.proto
```diff
@@ -0,0 +1,62 @@
...
+  int32 tx_power = 12;
+  bytes signature = 13;
+}
+
+message poc_lora_beacon_resp_v1 { string id = 1; }
+
+message poc_lora_witness_req_v1 {
+  bytes poc_key = 1;
+  bytes pub_key = 2;
+  bytes packet_hash = 3;
```
Vagabond, line 26: Use the full packet now, this way we could detect/account for bit errors in a way that a hash would not allow us to.

src/service/poc_lora.proto
```diff
@@ -0,0 +1,62 @@
...
+  bytes poc_key = 1;
+  bytes pub_key = 2;  
+  bytes entropy = 3;
+  bytes data = 4;
+  uint64 timestamp = 5;
+  sint32 signal = 6;
+  float snr = 7;
+  float frequency = 8;
+  int32 channel = 9;
+  string datarate = 10;
```
Vagabond, line 15: This can probably be an enum and avoid some pain we had with it being a string previously. Ask @mikev for the potential values

src/service/poc_lora.proto
```diff
@@ -0,0 +1,62 @@
...
+  bytes addr_hash = 11;
```
Vagabond, line 16: addr_hash is no longer necessary, I don't think.

src/service/poc_lora.proto
Vagabond: I think this should be the local entropy generated on the hotspot that is combined with the remote entropy to give a seed for generating the packet and the parameters.

src/service/poc_lora.proto
```diff
@@ -0,0 +1,62 @@
+syntax = "proto3";
+
+package helium.poc_lora;
+
+message poc_lora_beacon_req_v1 {
+  bytes poc_key = 1;
+  bytes pub_key = 2;  
+  bytes entropy = 3;
+  bytes data = 4;
+  uint64 timestamp = 5;
```
Vagabond, line 10: Several of these fields are going to have unknown values because this is a *transmit* claim, not a receive claim. Specifically timestamp (unless we can find a way to send a timed transmission without it being a packet response), signal and SNR are all unknown to the sender.

src/service/poc_lora.proto
```diff
@@ -0,0 +1,62 @@
...
+  bytes signature = 13;
+}
+
+message poc_lora_beacon_resp_v1 { string id = 1; }
+
+message poc_lora_witness_req_v1 {
+  bytes poc_key = 1;
+  bytes pub_key = 2;
+  bytes packet_hash = 3;
+  uint64 timestamp = 4;
```
Vagabond, line 27: the SX1302 and SX1303 have a higher resolution packet timer I'd like to collect the value from, if available. Either add a field to track the resolution of this value, or add a secondary field for the more precise time bits that are currently getting truncated.

src/service/poc_lora.proto
Vagabond: How can it be rejected?

src/service/poc_lora.proto
Vagabond: I'm confused what this proto will be used for, is this the result of the collation?

andymck: yes, this would be the result of the collation by the poc_verifier and would server as the rewardable POC

src/service/poc_lora.proto
andymck: well for example a witness could be rejected due to distance being too far etc.  The poc_verifier would be the entity performing the rejection based on any required validations

src/service/poc_lora.proto
andymck: added ts_res field to track resolution

src/service/poc_lora.proto
andymck: addressed here: https://github.com/helium/proto/pull/154/commits/69497c85fbe35ea4f5ba6d79155d9629b21854b8

src/service/poc_lora.proto
andymck: addressed here: https://github.com/helium/proto/pull/154/commits/69497c85fbe35ea4f5ba6d79155d9629b21854b8

src/service/poc_lora.proto
andymck: addressed here: https://github.com/helium/proto/pull/154/commits/69497c85fbe35ea4f5ba6d79155d9629b21854b8

src/service/poc_lora.proto
andymck: addressed here: https://github.com/helium/proto/pull/154/commits/69497c85fbe35ea4f5ba6d79155d9629b21854b8

src/service/poc_lora.proto
```diff
@@ -2,61 +2,69 @@ syntax = "proto3";
...
+enum DataRate {
+  SF7BW125 = 0;
+  SF8BW125 = 1;
+  SF9BW125 = 2;
+  SF10BW125 = 3;
+  SF12BW125 = 4;
+  // TODO: add final values
+}
+
+enum InvalidReason {
```
madninja, line 14: Don’t use capitalized names

src/service/poc_lora.proto
```diff
@@ -0,0 +1,70 @@
...
+  SF12BW125 = 4;
+  // TODO: add final values
+}
+
+enum invalid_reason {
+  bad_signature = 0;
+  // TODO: add final values
+}
+message beacon_report_req_v1 {
+  bytes beacon_id = 1;  // likely be a concatenation of local and remote entropy so maybe not needed if we already include both fields ?
```
Vagabond, line 19: this is a good question as we'd need to verify it was derived correctly anyway

Vagabond: I think it's ok to leave it here though

madninja: I'd remove this to avoid duplication?

src/service/poc_lora.proto
```diff
@@ -0,0 +1,72 @@
...
+  bad_signature = 0;
+  // TODO: add final values
+}
+message lora_beacon_report_req_v1 {
+  bytes beacon_id = 1;  // likely be a concatenation of local and remote entropy so maybe not needed if we already include both fields ? 
+  bytes pub_key = 2;  
+  bytes local_entropy = 3;
+  bytes remote_entropy = 4;
+  bytes data = 5;
+  float frequency = 6;
```
JayKickliter, line 24: units? If Hz, how about making it a u64?

src/service/poc_lora.proto
```diff
@@ -0,0 +1,72 @@
...
+  int32 channel = 7;
+  data_rate datarate = 8;
+  int32 tx_power = 9;
```
JayKickliter, line 27: units?

src/service/poc_lora.proto
```diff
@@ -0,0 +1,72 @@
...
+  // Timestamp of beacon report submit
+  uint64 timestamp = 10;
```
JayKickliter, line 29: UTC? Epoch?

src/service/poc_lora.proto
```diff
@@ -0,0 +1,72 @@
...
+package helium.poc_lora;
+
+enum data_rate {
+  SF7BW125 = 0;
+  SF8BW125 = 1;
+  SF9BW125 = 2;
+  SF10BW125 = 3;
+  SF12BW125 = 4;
+  // TODO: add final values
+}
```
JayKickliter, line 12: We should reuse exsiting proto def

dpezely:
FYI: ChirpStack v4 is implemented in Rust, and its repo contains protobufs and a lorawan library, `lrwn`.  It may be worthwhile to take a look:
- https://github.com/chirpstack/chirpstack.git
- `chirpstack/api/proto/` (identical to `chirpstack/api/rust/proto/chirpstack/`

src/service/poc_lora.proto
mikev: Please add all known DataRates, including RFU (I'd prefer we  make RFU as zero)

        'SF12BW125' -> {12, 125};
        'SF11BW125' -> {11, 125};
        'SF10BW125' -> {10, 125};
        'SF9BW125' -> {9, 125};
        'SF8BW125' -> {8, 125};
        'SF7BW125' -> {7, 125};
        'SF12BW250' -> {12, 250};
        'SF11BW250' -> {11, 250};
        'SF10BW250' -> {10, 250};
        'SF9BW250' -> {9, 250};
        'SF8BW250' -> {8, 250};
        'SF7BW250' -> {7, 250};
        'SF12BW500' -> {12, 500};
        'SF11BW500' -> {11, 500};
        'SF10BW500' -> {10, 500};
        'SF9BW500' -> {9, 500};
        'SF8BW500' -> {8, 500};
        'SF7BW500' -> {7, 500};
        'LRFHSS1BW137' -> {7, 125};
        'LRFHSS2BW137' -> {7, 125};
        'LRFHSS1BW336' -> {7, 125};
        'LRFHSS2BW336' -> {7, 125};
        'LRFHSS1BW1523' -> {7, 125};
        'LRFHSS2BW1523' -> {7, 125};
        'FSK50' -> {7, 125};
        'RFU' -> {7, 125};

src/service/poc_lora.proto
mikev: Normally in LoRa frequency is MHz.  I'd prefer float since it is 32 bits on the wire.

src/service/poc_lora.proto
```diff
@@ -0,0 +1,70 @@
...
+
+message lora_beacon_report_resp_v1 { string id = 1; }
+
+message lora_witness_report_req_v1 {
+  bytes beacon_id = 1;
+  bytes pub_key = 2;
+  bytes packet = 3;
+  uint64 timestamp = 4;
+  uint32 ts_res = 5;
+  sint32 signal = 6;
```
mikev, line 41: rssi?  rssi should be a uin32.

src/service/poc_lora.proto
```diff
@@ -0,0 +1,70 @@
...
+  float snr = 7;
+  float frequency = 8;
+  int32 channel = 9;
```
mikev, line 44: What is the purpose of channel?

src/service/service_core.proto
```diff
@@ -0,0 +1,38 @@
...
+    SF8BW500 = 16;
+    SF7BW500 = 17;
+    LRFHSS1BW137 = 18;
+    LRFHSS2BW137 = 19;
+    LRFHSS1BW336 = 20;
+    LRFHSS2BW336 = 21;
+    LRFHSS1BW1523 = 22;
+    LRFHSS2BW1523 = 23;
+    FSK50 = 24;
+    RFU = 25;
```
madninja, line 37: Remove RFU.. Reserver Future Use is not something we need to do in an enum

src/service/service_core.proto
```diff
@@ -0,0 +1,38 @@
+syntax = "proto3";
+
+package helium;
+
+enum gateway_staking_mode {
```
madninja, line 5: Can we just move these into individual files like `data_rate.proto` and `staking_mode.proto` to avoid the kitchen sink files syndrome?

src/service/service_core.proto
andymck: it was requested by mike above

src/service/service_core.proto
andymck: ev one needs a kitchen sink...but sure...

src/service/poc_lora.proto
```diff
@@ -0,0 +1,89 @@
...
+  lora_beacon_report_req_v1 beacon = 1;
+  invalid_reason reason = 2;
+}
+
+message lora_invalid_witness_report_v1 {
+  lora_witness_report_req_v1 witness = 1;
+  invalid_reason reason = 2;
+  // the participant to which the reason applies, which triggered the report as
+  // invalid
+  invalid_participant_type side = 3;
```
vihu, line 75:
```suggestion
  invalid_participant_type participant_type = 3;
```

src/service/poc_lora.proto
Vagabond: or use khz and make it a u32

src/service/poc_lora.proto
Vagabond: I think channel can be removed

src/service/poc_lora.proto
```diff
@@ -0,0 +1,91 @@
...
+}
+
+message lora_valid_poc_v1 {
+  bytes poc_id = 1;
+  lora_beacon_report_req_v1 beacon_report = 2;
+  repeated lora_witness_report_req_v1 witness_reports = 3;
+}
```
Vagabond, line 84: Can you add a signature field here?

src/service/poc_lora.proto
Vagabond: in other places we've done db*10 to avoid floats

src/service/poc_lora.proto
Vagabond: ideally this is, yeah, more precise than to the second

src/service/poc_lora.proto
Vagabond: Actually never mind on this.

src/service/poc_lora.proto
```diff
@@ -0,0 +1,91 @@
...
+
+message lora_ingest_report_v1 {
+  uint64 received_timestamp = 1;
+  oneof report {
+    lora_beacon_report_req_v1 beacon = 2;
+    lora_witness_report_req_v1 witness = 3;
+  }
+}
+
+message lora_beacon_report_req_v1 {
```
Vagabond, line 32: Why is there a `req` in the name here?

andymck: the 'req' protos are the incoming grpc protos....ie the gateway rs client submits requests of type: lora_beacon_report_req_v1 and lora_witness_report_req_v1.  The ingestor writes out those req protos.

The verifier then consumes those protos and writes them out as part of its valid and invalid protos.  The one/same proto is used as there is an existing encoder/decoder capability.

The alternative is to have a new proto but you will end up with the same structure and thus duplication

src/service/poc_lora.proto
```diff
@@ -0,0 +1,91 @@
...
+  bytes pub_key = 2;
+  bytes packet = 3;
+  uint64 timestamp = 4;
+  uint32 ts_res = 5;
+  sint32 signal = 6;
+  float snr = 7;
+  float frequency = 8;
+  int32 channel = 9;
+  data_rate datarate = 10;
+  bytes signature = 11;
```
vihu, line 62: We also need to attach location to the witness and beacon reports. So when the verifier is outputting the files to S3, it needs to query the follower and attach location to the witness and beacon reports.

Furthermore, we also need a place for attaching the hex_scale (or tx_scale as its specified in rewards_v2 on core). So that the injector service can just look up the files output from verifier directly and just construct the poc_receipts_v2 transaction.

src/service/poc_lora.proto
```diff
@@ -0,0 +1,123 @@
...
+}
+
+// response returned to gateway submitting witness report to ingestor
+message lora_witness_report_resp_v1 { string id = 1; }
+
+// beacon report outputted by ingestor, tagged with received timestamp
+message lora_beacon_ingest_report_v1 {
+  uint64 received_timestamp = 1;
+  lora_beacon_report_req_v1 report = 2;
```
vihu, line 68: Should stick to one naming convention imo. So like:
`lora_beacon_report_req_v1 beacon_report` or `lora_beacon_report_req_v1 beacon`
`lora_witness_report_req_v1 witness_report` or `lora_witness_report_req_v1 witness`

throughout the proto

andymck: updated to use 'report': https://github.com/helium/proto/pull/154/commits/a7bd1184cf9e289dfb8aee7cbf865f1ece05d77d

src/service/poc_lora.proto
andymck: switched frequency to uint32. cc7a2e1644eeeae1d54d94e70e35611f3ea459e5

src/service/poc_lora.proto
andymck: removed here: cc7a2e1644eeeae1d54d94e70e35611f3ea459e5

src/service/poc_lora.proto
andymck: addressed here: cc7a2e1644eeeae1d54d94e70e35611f3ea459e5

src/service/poc_lora.proto
andymck: channel removed here: cc7a2e1644eeeae1d54d94e70e35611f3ea459e5

madninja (approved):

PR #157, opened 2022-08-05
Packet Router

macpie:


src/service/packet_router_service.proto
```diff
@@ -0,0 +1,10 @@
+syntax = "proto3";
+
+package helium;
+
+import "packet_router.proto";
```
madninja, line 5: so far we've been inlining the messages with the service. You mind inlining packet_router.proto?

madninja: oh and run the formatter on it :-)

macpie: Will do I keep it outside right now cause we don't have grpc implemented yet

madninja: doesn't matter.. you can still use the messages defined in a service even if you're not done with the service yet

macpie: it seems to not compile the `.hrl` file

madninja: gah gpb madness... presumably @andymck has had this working just fine

src/service/packet_router.proto
```diff
@@ -0,0 +1,60 @@
...
+  packet_router_route_protocol_v1 protocol = 5;
+  uint64 oui = 6;
+}
+
+message packet_router_packet_up_v1 {
+  bytes payload = 1;
+  uint64 timestamp = 2;
+  float signal_strength = 3;
+  float frequency = 4;
+  string datarate = 5;
```
madninja, line 37: Use the same data rate that @andymck is using in his PR?

src/service/packet_router.proto
```diff
@@ -0,0 +1,60 @@
...
+  region region = 7;
+  uint64 hold_time = 8;
+  bytes hotspot = 9;
+  bytes signature = 10;
+}
+
+message window_v1 {
+  uint64 timestamp = 1;
+  float frequency = 2;
+  string datarate = 3;
```
madninja, line 48: Same

src/service/packet_router.proto
```diff
@@ -0,0 +1,87 @@
...
+message packet_router_packet_up_v1 {
+  bytes payload = 1;
+  uint64 timestamp = 2;
+  uint32 rssi = 3;
+  float frequency_mhz = 4;
+  data_rate datarate = 5;
+  float snr = 6;
+  region region = 7;
+  uint64 hold_time = 8;
+  bytes gateway = 9;
```
dpezely, line 48: The `gateway` field may need to include more information than just bytes of its literal key.  For instance, see level of detail in this Rust library:
https://github.com/helium/helium-crypto-rs/blob/main/src/public_key.rs

Namely, which environment? mainnet, testnet; which key type? ECC Compact, ed25519; etc.

While we don't necessarily need to track mainnet/testnet in this field, it should be captured somewhere.

src/data_rate.proto
```diff
@@ -0,0 +1,31 @@
+syntax = "proto3";
+
+package helium;
+
+enum data_rate {
+  SF12BW125 = 0;
+  SF11BW125 = 1;
```
mikev, line 7: Google recommends UNSPECIFIED for the 0 default.  I'd recommend the same.
It can happen that the DataRate is unknown or unspecified.  SF12BW125 is the standard default for most plans so that might work for most plans, but SF12BW125 is an illegal DataRate for US915.

https://developers.google.com/protocol-buffers/docs/proto3#enum

dpezely:
Superseded by https://github.com/helium/proto/pull/193

PR #158, opened 2022-08-05
add region_params

madninja:


src/service/gateway.proto
```diff
@@ -51,6 +51,13 @@ message gateway_region_params_update_req_v1 {
   bytes signature = 2;
 }
 
+message gateway_region_params_req_v1 { region region = 1; }
+message gateway_region_params_resp_v1 {
+  region region = 2;
```
andymck, line 56: ordering is starting at 2

madninja: Fixed in 3a9dde1

src/service/gateway.proto
```diff
@@ -51,6 +51,13 @@ message gateway_region_params_update_req_v1 {
   bytes signature = 2;
 }
 
+message gateway_region_params_req_v1 { region region = 1; }
+message gateway_region_params_resp_v1 {
+  region region = 1;
+  blockchain_region_params_v1 params = 2;
+  uint64 gain = 3;
```
andymck, line 58: the problem here is that the 'gain' value is derived from an gateway data, which itself is set at the point the GW is asserted.  Not sure there is much value in adding this field, if we have to default it then the client may as well do this itself

src/service/gateway.proto
```diff
@@ -70,6 +77,7 @@ message gateway_resp_v1 {
     gateway_poc_check_challenge_target_resp_v1 poc_check_target_resp = 17;
     gateway_public_routing_data_resp_v1 public_route = 18;
     gateway_version_resp_v1 version = 19;
+    gateway_region_params_resp_v1 region_params__resp = 20;
```
andymck, line 80: there is an exta or spurious underscore between params and resp

src/service/gateway.proto
madninja: not sure what you mean here.. This is the same response body as the stream response? What's different for a unary request?

andymck: With the unary request, no gateway is being specified.  The 'gain' value returned in the streaming request is derived from the gateway info, with the gateway itself being specified in the request.

andymck: https://github.com/helium/sibyl/blob/master/src/grpc/helium_stream_region_params_update_impl.erl#L138

madninja: hmm, so we need that gain to adjust the tx power on join. I guess I should pass in the gateway key? We can also sign the request to avoid mass querying?

michaeldjeffrey (approved):

evanmcc (approved):

PR #162, opened 2022-08-15
Add packet reporting message to packet router

hwahby:
Draft for protobuff packet data uploaded to S3


src/service/packet_router.proto
```diff
@@ -48,6 +48,17 @@ message packet_router_packet_up_v1 {
   bytes signature = 10;
 }
 
+message packet_router_packet_report_v1 {
+  bytes packet_hash = 1;
+  uint64 timestamp = 2;
+  uint32 oui = 3;
+  float signal_strength = 4;
+  float frequency = 5;
+  string datarate = 6;
```
madninja, line 57: Use an actual data rate type

hwahby: I'm seeing data rate as type `string` elsewhere in the repo and no `datarate` type. We can create a new type. What would be the parameters? A string of numbers? Number of bits + units (i.e 50kbs)?

madninja: Coordinate with @andymck here but I think we should have a proto definition as defined here: https://github.com/helium/proto/pull/154/files#diff-6c154ef0b4726a72a348393855c692f17dc215d4201ab43b2660bb38dac0b6b3R5-R12 to have actual types for Datarate

macpie: @mikev might be able to help here as well

mikev: I agree datarate should be an enum.  An enum only uses one byte on disk or over the wire.

Here are the current known values:

        'SF12BW125' -> {12, 125};
        'SF11BW125' -> {11, 125};
        'SF10BW125' -> {10, 125};
        'SF9BW125' -> {9, 125};
        'SF8BW125' -> {8, 125};
        'SF7BW125' -> {7, 125};
        'SF12BW250' -> {12, 250};
        'SF11BW250' -> {11, 250};
        'SF10BW250' -> {10, 250};
        'SF9BW250' -> {9, 250};
        'SF8BW250' -> {8, 250};
        'SF7BW250' -> {7, 250};
        'SF12BW500' -> {12, 500};
        'SF11BW500' -> {11, 500};
        'SF10BW500' -> {10, 500};
        'SF9BW500' -> {9, 500};
        'SF8BW500' -> {8, 500};
        'SF7BW500' -> {7, 500};
        'LRFHSS1BW137' -> {7, 125};
        'LRFHSS2BW137' -> {7, 125};
        'LRFHSS1BW336' -> {7, 125};
        'LRFHSS2BW336' -> {7, 125};
        'LRFHSS1BW1523' -> {7, 125};
        'LRFHSS2BW1523' -> {7, 125};
        'FSK50' -> {7, 125};
        'RFU' -> {7, 125};

src/service/packet_router.proto
```diff
@@ -48,6 +48,17 @@ message packet_router_packet_up_v1 {
   bytes signature = 10;
 }
 
+message packet_router_packet_report_v1 {
+  bytes packet_hash = 1;
+  uint64 timestamp = 2;
+  uint32 oui = 3;
+  float signal_strength = 4;
```
mikev, line 55: replace with rssi.  The name rssi is customarily used in LoRa documents.

src/service/packet_router.proto
```diff
@@ -48,6 +48,17 @@ message packet_router_packet_up_v1 {
   bytes signature = 10;
 }
 
+message packet_router_packet_report_v1 {
+  bytes packet_hash = 1;
```
mikev, line 52: replace with packet

src/data_rate.proto
```diff
@@ -0,0 +1,32 @@
...
+    SF8BW500 = 16;
+    SF7BW500 = 17;
+    LRFHSS1BW137 = 18;
+    LRFHSS2BW137 = 19;
+    LRFHSS1BW336 = 20;
+    LRFHSS2BW336 = 21;
+    LRFHSS1BW1523 = 22;
+    LRFHSS2BW1523 = 23;
+    FSK50 = 24;
+    RFU = 25;
```
dpezely, line 31: fRFU is "reserved for future use", and Michael asked that this be assigned to the 0th slot, which of course upsets the numbering

```suggestion
    RFU = 0;
```

hwahby: Ah, right, will adjust.

madninja: not quite how protobuf is intended to be used. Remove RFU.. protobuf messages can be extended for future use and you shouldn't be assigning index

src/service/packet_router.proto
```diff
@@ -44,10 +45,23 @@ message packet_router_packet_up_v1 {
...
+  uint64 gateway_timestamp_ms = 1;
+  uint32 oui = 2;
+  uint64 net_id = 3;
+  uint32 rssi = 4;               // LoRa signal strength
+  float frequency_mhz = 5;
+  float snr = 6;
+  data_rate datarate = 7;
+  region region = 8;
+  bytes gateway = 9;
+  bytes lora_payload_hash = 10;  // Hash of `payload` within `message packet`
```
madninja, line 62: remove to payload_hash.. otherwise endless arguments as to whether it's lora or lorawan payload :-/

src/service/packet_router.proto
```diff
@@ -60,7 +74,30 @@ message packet_router_packet_down_v1 {
   window_v1 rx2 = 3;
 }
 
-service hotspot {
+
```
madninja, line 77: I don't think the verifier report is needed and as mentioned in discord I'm not convinced yet that linking is quite the way to get easy auditing/reporting.

dpezely: Removed with commit 43d090e

src/data_rate.proto
```diff
@@ -3,29 +3,29 @@ syntax = "proto3";
...
-    SF8BW500 = 17;
-    SF7BW500 = 18;
-    LRFHSS1BW137 = 19;
-    LRFHSS2BW137 = 20;
-    LRFHSS1BW336 = 21;
-    LRFHSS2BW336 = 22;
-    LRFHSS1BW1523 = 23;
-    LRFHSS2BW1523 = 24;
-    FSK50 = 25;
+    SF12BW125 = 0;
```
madninja, line 6: One more time. Field numbers start at 1

hwahby: Reverted per: https://developers.google.com/protocol-buffers/docs/proto3#default

"every enum definition must contain a constant that maps to zero as its first element."

madninja: No. that's the_encoded_ value in language mapping. I shared this last night already <https://developers.google.com/protocol-buffers/docs/proto3#assigning_field_numbers>

hwahby: Is there a difference in numbering for messages vs enums? My IDE was displaying an error when I removed the 0 count, which made me take a second look.

michaeldjeffrey: double/triple checked. Enums values have to have a `0`.

michaeldjeffrey:
Merging
The messages this PR was trying to address are in a usable state.
More updates should be more PRs with their own reasons instead of piling on this one.
And we're still going into another branch that hasn't been merged yet.

src/data_rate.proto
madninja: what.. do we do this everywhere? I must've blanked this from my memory if so

madninja: I clearly did.. apologies <https://developers.google.com/protocol-buffers/docs/proto3#enum> specs the enum requirements

PR #163, opened 2022-08-15
remove optional parameters

jeffgrunewald:
optional parameters don't play nice everywhere we need them to so these all need to be "required" now

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,3 +1,4 @@
+#![allow(clippy::derive_partial_eq_without_eq)]
 include!(concat!(env!("OUT_DIR"), "/helium.rs"));
 
 pub use blockchain_txn::Txn;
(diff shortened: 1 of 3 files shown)


madninja (changes requested):
Looks like clipoy is angry 
https://github.com/helium/proto/runs/7845965174?check_suite_focus=true#step:7:76

jeffgrunewald:
> Looks like clipoy is angry https://github.com/helium/proto/runs/7845965174?check_suite_focus=true#step:7:76

According to various issues in the community, this is debatably a bug in clippy; we don't pin the version of rust we run on CI and rust 1.63.0 was just released two days ago. Guess which version of rust this happens with? There are also opinions that enforcing deriving `Eq` when you're deriving `PartialEq` constrains library development and shouldn't be a linting error.

Given the uncertainty I'm inclined to throw an `ignore this rule` config in and move along but I'm open to suggestions

PR #165, opened 2022-08-17
Do not use protobuf-src for Windows builds

JayKickliter:
protobuf-src appears to be broken on windows. The problem we're solving by using protobuf-src only applies to Linux, so let's use system protoc for Windows.


madninja (approved):

madninja:
Hell of a hack!

PR #167, opened 2022-08-31
Add verifier protobuf

maplant:


vihu:
Seems like a decent start, though I wonder if we need the `map` type specifically for some of the shares here? Can that not just be constructed in rust/erlang if we have the underlying proto?

Secondly, we should add something like `invalid_data_reason` or `unrewardable_shares`, specifically capturing the data being skipped [here](https://github.com/novalabsxyz/poc5g-server/blob/main/rewards/src/reward_share.rs#L170). So we can have `hotspot_pubkey`, `timestamp`, `after`, `before`, `cbsd_id` and `operation_mode` to begin with I suppose

src/service/verifier.proto
```diff
@@ -0,0 +1,92 @@
...
+message speed_share_list {
+  repeated speed_share speed_shares = 1;
+}
+
+message speed_shares {
+  map<string, speed_share_list> speed_shares = 1;
+}
+
+message moving_avg {
+  uint64 window_size = 1;
```
JayKickliter, line 44: A comment about units would helpful. Same for any other applicable fields in the new protos

src/service/verifier.proto
```diff
@@ -0,0 +1,92 @@
...
+  average average = 2;
+  bool is_valid = 3; 
+  repeated speed_share speed_shares = 4;
+}
+
+message average {
+  uint64 upload_speed_avg = 1;
```
JayKickliter, line 51: Another good case for units

maplant:
> Seems like a decent start, though I wonder if we need the `map` type specifically for some of the shares here? Can that not just be constructed in rust/erlang if we have the underlying proto?

We could reconstruct it, but that seems more difficult. I think that anything we can use to map directly to rust types seems like a positive in that it will reduce the amount of errors we see.

src/service/poc_mobile.proto
```diff
@@ -38,4 +38,133 @@ service poc_mobile {
...
+}
+
+enum file_type {
+  heartbeat = 0;
+  speedtest = 1;
+}
+
+enum cell_type {
+  nova436h = 0;
+  nova4301 = 1;
```
vihu, line 166:
```suggestion
  nova430i = 1;
```

src/service/poc_mobile.proto
```diff
@@ -38,4 +38,133 @@ service poc_mobile {
...
+message moving_avg {
+  bytes pub_key = 1;
+  uint64 window_size = 2;
+  average average = 3;
+  bool is_valid = 4; 
+  repeated speed_share speed_shares = 5;
+}
+
+message average {
+  uint64 upload_speed_avg = 1;
```
vihu, line 105:
```suggestion
  /// bytes/sec
  uint64 upload_speed_avg = 1;
```

src/service/poc_mobile.proto
```diff
@@ -38,4 +38,133 @@ service poc_mobile {
...
+  uint64 download_speed_avg =  2;
```
vihu, line 106:
```suggestion
  /// bytes/sec
  uint64 download_speed_avg =  2;
```

src/service/poc_mobile.proto
```diff
@@ -38,4 +38,133 @@ service poc_mobile {
...
+  uint32 latency_avg = 3;
```
vihu, line 107:
```suggestion
  /// in ms
  uint32 latency_avg = 3;
```

src/service/poc_mobile.proto
```diff
@@ -38,4 +38,133 @@ service poc_mobile {
...
+}
+
+message speed_share_moving_avgs {
+  repeated moving_avg moving_avgs = 1; 
+}
+
+message cell_share {
+  cell_type cell_type = 1;
+  uint64 weight = 2;
```
vihu, line 116:
```suggestion
  /// weight will be x10, i.e. 40 = 4.0
  uint64 weight = 2;
```

src/service/poc_mobile.proto
```diff
@@ -38,4 +38,133 @@ service poc_mobile {
...
+  string invalid_reason = 6;
+}
+
+message unrewardable_shares {
+  repeated unrewardable_share shares = 1;
+}
+
+message speed_share {
+  bytes pub_key = 1;
+  uint64 upload_speed = 2;
```
vihu, line 81: Add units here as well please.

src/service/poc_mobile.proto
```diff
@@ -38,4 +38,132 @@ service poc_mobile {
...
+}
+
+message speed_share_list {
+  bytes gw_pub_key = 1;
+  repeated speed_share speed_shares = 2;
+}
+
+message speed_shares { repeated speed_share_list speed_shares = 1; }
+
+message invalid_speed_share {
```
jeffgrunewald, line 90: do we need both valid and invalid types for speed shares when the only difference is the addition of the `reason` string on the invalid side?
any reason not to combine them, possibly with a `is_valid` boolean field?

src/service/poc_mobile.proto
```diff
@@ -38,4 +38,132 @@ service poc_mobile {
...
+  bytes pub_key = 1;
+  /// bytes/sec
+  uint64 upload_speed = 2;
+  /// bytes/sec
+  uint64 download_speed = 3;
+  uint32 latency = 4;
+  uint64 timestamp = 5;
+  string invalid_reason = 6;
```
jeffgrunewald, line 98: do we need a string here or is this something we can reasonably enumerate like an http error?

maplant: I think we can enumerate it but I'm not sure that I can by myself. The only invalid reason I found was "not in time range"

JayKickliter (changes requested):
Thinking more about units, I'm leaning towards having them appended to field names, like `latency_average_ms` or `download_speed_avg_bps`. But if we do, let's be consistent about it.

CC @vihu

src/service/poc_mobile.proto
maplant: I think both ways are valid and it just comes down to personal preference. However, I will say that we are likely to only put the valid protos in buckets that will be inputs to other services, so I'd prefer those services to not have to potentially deal with "is_invalid" as a required false value

jeffgrunewald: that depends on what you consider "other services"; we'll need to expose reports of both valid _and_ invalid reward inputs for accountability so _some_ services will be consuming the invalid speed shares and heartbeats even if it's not the rewards servers

src/service/poc_mobile.proto
vihu: > I think we can enumerate it but I'm not sure that I can by myself. The only invalid reason I found was "not in time range"

Well, there's definitely multiple invalid reasons, some being:
- heartbeat decode failed for some reason, although ingester service might not allow that cc: @madninja 
- gateway owner not found
- heartbeat not within time range
- unsupported cbsd_id -> cell_type

vihu:
> Thinking more about units, I'm leaning towards having them appended to field names, like `latency_average_ms` or `download_speed_avg_bps`. But if we do, let's be consistent about it.
> 
> CC @vihu

I'm fine with that, @maplant can you make the required changes please, basically tag the units in the field name itself. And we need to ensure that the changes would be reflected in https://github.com/novalabsxyz/poc5g-server/pull/49 as well.

src/service/poc_mobile.proto
maplant: I'll go ahead and combine as many as I can - I do need to give invalid_speed_shares as a separate list because they may not have a valid gw_pub_key

src/service/poc_mobile.proto
```diff
@@ -38,4 +38,111 @@ service poc_mobile {
...
+
+message shares { repeated share shares = 1; }
+
+message speed_share {
+  bytes pub_key = 1;
+  /// bytes/sec
+  uint64 upload_speed_bps = 2;
+  /// bytes/sec
+  uint64 download_speed_bps = 3;
+  uint32 latency = 4;
```
JayKickliter, line 69: unites please. is it ms?
```suggestion
  uint32 latency_ms = 4;
```

JayKickliter (changes requested):

JayKickliter (approved):

PR #169, opened 2022-09-02
implement FromStr for DataRate enum

michaeldjeffrey:
Questions:
- Is DecodeError implemented properly?
- An easier way to compare strings?
- Is the impl block scoped correctly?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -75,7 +75,7 @@ pub mod services {
                 "LRFHSS1BW1523" => Ok(DataRate::Lrfhss1bw1523),
                 "LRFHSS2BW1523" => Ok(DataRate::Lrfhss2bw1523),
                 "FSK50" => Ok(DataRate::Fsk50),
-                unknown => Err(prost::DecodeError::new(format!("unknown datarate: {}", unknown)))
+                unknown => Err(prost::DecodeError::new(format!("unknown datarate: {unknown}")))
             }
         }
     }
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -43,4 +44,39 @@ pub mod services {
...
+                "SF8BW500" => Ok(DataRate::Sf8bw500),
+                "SF7BW500" => Ok(DataRate::Sf7bw500),
+                "LRFHSS1BW137" => Ok(DataRate::Lrfhss1bw137),
+                "LRFHSS2BW137" => Ok(DataRate::Lrfhss2bw137),
+                "LRFHSS1BW336" => Ok(DataRate::Lrfhss1bw336),
+                "LRFHSS2BW336" => Ok(DataRate::Lrfhss2bw336),
+                "LRFHSS1BW1523" => Ok(DataRate::Lrfhss1bw1523),
+                "LRFHSS2BW1523" => Ok(DataRate::Lrfhss2bw1523),
+                "FSK50" => Ok(DataRate::Fsk50),
+                unknown => Err(prost::DecodeError::new(format!("unknown datarate: {}", unknown)))
```
madninja, line 78: You can put the unknown variable in the placeholder of the format call

madninja:
Technically the impl doesn’t have to be in the services section at all but it doesn’t hurt either

madninja:
There are tactics for a perfect parser given the known set but this is fine for now I think

src/lib.rs
michaeldjeffrey: Tried that before I added `format!` and forgot to try it again when I did. Thank you

PR #172, opened 2022-09-09
Rename validity, make invalid optional type for shares, add helper

maplant:


src/service/poc_mobile.proto
```diff
@@ -55,7 +55,7 @@ message share {
   uint64 weight = 3;
   uint64 timestamp = 4;
   cell_type cell_type = 5;
-  validity validity = 6;
+  optional invalid_share_reason invalid = 6;
```
JayKickliter, line 58: we recently removed all optional fields in our protos. I think because it's a pain keeping CI systems up to date with a newish `protoc`

maplant: I think that we can use a `oneof` to get the same behavior

maplant: Updated with a version that uses oneof instead of optional

madninja (approved):

PR #175, opened 2022-09-13
Mapper Protos

lthiery:


madninja (commented):
You're going to want to add them to a build.rs feature to get picked up by users of the crate?

lthiery:
Will do. I'm just trying to figure out the design of these first though. I'm wondering what to do about the signer's pubkey. Is it important that it be part of what's signed or just that it ends up in the final stored proto for verification?

src/service/attach_response.proto
```diff
@@ -0,0 +1,37 @@
+syntax = "proto3";
+package helium.mapper;
+
+import "service/gps_response.proto";
+
+message attach_response_t {
```
madninja, line 6: Can we drop the `_t` convention throughout this PR?

src/service/gps_response.proto
```diff
@@ -0,0 +1,23 @@
+syntax = "proto3";
+
+package helium.mapper;
+
+message gps_response_t {
+  oneof version { gps_response_v1 gps_response_v1 = 1; }
+}
+
+message gps_response_v1 {
+  /**<   Unix time in seconds */
```
madninja, line 10: Let's use the same `//` comment style throughout please

madninja (commented):
Looks pretty good! Some stylistic comments

src/service/attach_response.proto
lthiery: Yeah, I liked it at first because it made conventional C-named struct but it's gross elsewhere.

src/service/gps_response.proto
lthiery: You got it

sabondano:
Would it make sense to some sort of device/mapper identifier to scan and attach protos?

lthiery:
> Would it make sense to add some sort of device/mapper identifier to scan and attach protos? Maybe it's there and I'm not seeing it.

It's not there. 

> Would it make sense to add some sort of device/mapper identifier to scan and attach protos? Maybe it's there and I'm not seeing it.

Good point. I added a signer field. In some situations, we will need this to be populated by whoever is parsing the LoRaWAN frame as it will add unnecessary bloat to the LoRaWAN frame.

src/service/modem_response.proto
```diff
@@ -5,11 +5,12 @@ import "service/attach_response.proto";
...
 
-message modem_response_t {
+message modem_response {
   oneof message {
     gps_response gps_response = 1;
     scan_response scan_response = 2;
     attach_response attach_response = 3;
   }
   repeated bytes signature = 4;
+  repeated bytes signer = 5;
```
madninja, line 15: uhh.. thus us called `pub_key` everywhere else and would be needed to verify the signature anyway

sabondano:
I was also wondering if we should add a comment that describes what exactly is signed 🤔 , for others to know how to verify it.

src/service/modem_response.proto
lthiery:
```
message blockchain_signature_v1 {
  bytes signer = 1;
  bytes signature = 2;
}
```

but if we prefer the pub_key / signature convention, that's fine.

lthiery:
@sabondano i added a comment about what is signed. let me know if that's clear now

src/service/modem_response.proto
madninja: yeah just going by what the newer protos for all the grpc services do there. 

Thanks for adjusting

madninja (approved):

PR #177, opened 2022-09-13
Add service definitions for Helium Config Service

barndon:
This PR adds `service/config.pb` which contains the API for Helium Config Service.

### DIFF
--- a/src/service/config.proto
+++ b/src/service/config.proto
@@ -1,6 +1,6 @@
 syntax = "proto3";
 
-package helium_config;
+package helium.config;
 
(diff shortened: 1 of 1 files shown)


src/service/config.proto
```diff
@@ -0,0 +1,36 @@
+syntax = "proto3";
+
+package helium_config;
```
jeffgrunewald, line 3:
```suggestion
package helium.config;
```
The convention we’ve followed so far is dot-separating the elements of the package name for services

PR #179, opened 2022-09-15
Add subnetwork_rewards message

maplant:
Either one of these is fine for me

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -9,10 +9,7 @@ fn main() -> Result<()> {
     tonic_build::configure()
         .build_server(true)
         .build_client(true)
-        .type_attribute(
-            ".",
-            "#[derive(serde_derive::Serialize, serde_derive::Deserialize)]",
-        )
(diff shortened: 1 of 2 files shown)


build.rs
```diff
@@ -9,7 +9,10 @@ fn main() -> Result<()> {
     tonic_build::configure()
         .build_server(true)
         .build_client(true)
-        .type_attribute(".", "#[derive(serde_derive::Serialize)]")
+        .type_attribute(
+            ".",
+            "#[derive(serde_derive::Serialize, serde_derive::Deserialize)]",
```
madninja, line 14: this may break downstream users of the crate like api-rs (not tried yet). You don't need this derive for what this PR intended right?

maplant: Yeah I don't, thought that it would be a pretty simple change but if it's not I'll get rid of it

madninja: thanks yeah.. just an unknown so would need some testing with a few key downstream crates

madninja (approved):

PR #183, opened 2022-09-19
Added report message that wraps speed test and hearbeat with a received timestamps

bbalser:


src/service/poc_mobile.proto
```diff
@@ -37,6 +43,12 @@ message cell_heartbeat_req_v1 {
 
 message cell_heartbeat_resp_v1 { string id = 1; }
 
+//cell heartbeat report outputted by ingestor, tagged with received_timestamp
+message cell_heartbeat_ingest_report_v1 {
+    uint64 received_timestamp = 1;
```
madninja, line 48: annotate with the expected unit for timestamp

src/service/poc_mobile.proto
```diff
@@ -37,6 +43,12 @@ message cell_heartbeat_req_v1 {
 
 message cell_heartbeat_resp_v1 { string id = 1; }
 
+//cell heartbeat report outputted by ingestor, tagged with received_timestamp
```
madninja, line 46: Run cargo fmt.. or enable format on save in your editor

src/service/poc_mobile.proto
bbalser: Fixed b07a1d1

src/service/poc_mobile.proto
bbalser: Ran `cargo fmt` on the project and it is not showing any changes, but still failing

madninja: oh my bad.. cargo fmt is for rust code.. you need to install clang-format and run the proto formatter that it supports.  vscode has a plugin that does this work for you if you use that

madninja (approved):

PR #184, opened 2022-09-21
Use hz for frequencies

madninja:


andymck (approved):

src/service/poc_lora.proto
```diff
@@ -28,8 +28,8 @@ message lora_beacon_report_req_v1 {
   bytes local_entropy = 3;
   bytes remote_entropy = 4;
   bytes data = 5;
-  // Frequency in kHz
-  uint32 frequency = 6;
+  // Frequency in Hz
+  uint64 frequency = 6;
```
jeffgrunewald, line 32: `frequency_hz`

madninja: > frequency_hz

Really? We haven't used that convention anywhere else.

jeffgrunewald: @maplant took the suggestion in some of his updates to the poc_mobile proto such as https://github.com/helium/proto/blob/571b71f48faf45ce5d09a3245aa9a782b6fb2a6d/src/service/poc_mobile.proto#L86; i don't know whether or not it makes any difference with type safety but it at least helps with clarity of the scale you're working with without having to reference the source. i imagine with language servers auto-completing fields you can see right away whether or not you need to convert the value you've got or if it's already in the units you expect

JayKickliter: We don't have a convention unfortunately. I guess it's technically a deci deci bel

madninja: so perhaps that'd be `gain_ddb`?

jeffgrunewald: That makes sense to me. And that would be a good opportunity for a clarifying comment

PR #186, opened 2022-09-22
Add version to entropy

madninja:
The version (default to 0) indicates how the entropy is intended to be decoded and used.

The current default version is 0. It indicates that the beacon payload is to be constructed using the sha256 of the remote entropy, remote entropy timestamp, and local entropy in that order.

This is implemented in the `entropy` workspace member of the gateway-rs repo

### DIFF
--- a/src/entropy.proto
+++ b/src/entropy.proto
@@ -8,4 +8,6 @@ message entropy {
   /// The start time for the validity of this entropy in seconds since unix
   // epoch
   uint64 timestamp = 2;
+  /// The entro[y version
+  uint32 version = 3;
 }
--- a/src/entropy.proto
+++ b/src/entropy.proto
@@ -2,7 +2,7 @@ syntax = "proto3";
 
 package helium;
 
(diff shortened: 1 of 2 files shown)


src/entropy.proto
```diff
@@ -8,4 +8,6 @@ message entropy {
   /// The start time for the validity of this entropy in seconds since unix
   // epoch
   uint64 timestamp = 2;
+  /// The entro[y version
```
andymck, line 11:
```suggestion
  /// The entropy version
```

andymck (approved):

andymck:
Should we rename the proto entropy_v1 or entropy_report_v1 to stick with convention of the other protos ?  Would distinguish it better from an internal rust struct

madninja:
> Should we rename the proto entropy_v1 or entropy_report_v1 to stick with convention of the other protos ? Would distinguish it better from an internal rust struct

Fixed in fcf2f50

PR #187, opened 2022-09-22
Add start and end epoch to subnetwork rewards

maplant:


src/blockchain_txn_subnetwork_rewards_v1.proto
```diff
@@ -8,7 +8,11 @@ message subnetwork_reward {
   uint64 amount = 2;
 }
 
-message subnetwork_rewards { repeated subnetwork_reward rewards = 1; }
+message subnetwork_rewards {
```
jeffgrunewald, line 11: can we safely reorder these? assuming this proto is only used in the verifier so far (and thus not running anywhere yet) i'll approve conditionally

jeffgrunewald (approved):

src/blockchain_txn_subnetwork_rewards_v1.proto
maplant: Yeah, this hasn't been used in anything besides the verifier. (I wouldn't modify this otherwise)

PR #190, opened 2022-09-28
Allow batching of gw info results

jeffgrunewald:
allow streaming active gateway results in batches instead of one at a time since there are hundreds of thousands of records at a given time

### DIFF
--- a/src/service/follower.proto
+++ b/src/service/follower.proto
@@ -25,9 +25,6 @@ message follower_gateway_req_v1 {
   bytes address = 1;
 }
 
-/// Request a stream of all active gateways from the on-chain metadata
-message follower_gateway_stream_req_v1 {}
-
 message follower_gateway_resp_v1 {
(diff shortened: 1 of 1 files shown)


src/service/follower.proto
```diff
@@ -25,9 +25,6 @@ message follower_gateway_req_v1 {
   bytes address = 1;
 }
 
-/// Request a stream of all active gateways from the on-chain metadata
-message follower_gateway_stream_req_v1 {}
```
madninja, line 29: Add Bach size as a request parameter? constrain to max. < 5000 or something?

jeffgrunewald: good call; added [17fc5ff](https://github.com/helium/proto/pull/190/commits/17fc5ff2fd774d07c934d682ce4eb68124a6c699)

madninja (approved):

PR #193, opened 2022-09-30
Add config, packet_router protos

dpezely:
This supersedes https://github.com/helium/proto/pull/157 which was too convoluted due to many `merge master`, etc.

Add:
- `package helium.config`
- `package helium.packet_router`

### DIFF
--- a/src/service/packet_router.proto
+++ b/src/service/packet_router.proto
@@ -1,0 +1,54 @@
+syntax = "proto3";
+
+package helium.packet_router;
+
+import "region.proto";
+import "data_rate.proto";
+
+message packet_router_packet_up_v1 {
+  bytes payload = 1;
+  uint64 timestamp = 2;
+  uint32 rssi = 3;
+  // Frequency in hz
+  uint32 frequency = 4;
+  data_rate datarate = 5;
+  float snr = 6;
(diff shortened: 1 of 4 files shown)


michaeldjeffrey (approved):

src/service/packet_router.proto
```diff
@@ -0,0 +1,54 @@
...
+  data_rate datarate = 3;
+}
+
+message packet_router_packet_down_v1 {
+  bytes payload = 1;
+  window_v1 rx1 = 2;
+  window_v1 rx2 = 3;
+}
+
+service gateway {
```
madninja, line 51: why is this called `gateway`? it's the `packet_router` service no?

michaeldjeffrey: I remember that it's gone through quite a few name changes. 
If I'm reaching for plausible reasons how we arrived here...

In the Packet Router namespace, this is the service that Gateways use.

macpie: yes I think that the only real reason

macpie: We changed it to `packet` as this is what the service is moving around

madninja (approved):

PR #194, opened 2022-10-03
add secure_packet_v1 for secure concentrators

dinocore1:
Secure LoRaWAN Concentrator cards sign RX packets in hardware using unique ED25519 key. The signature consists of the actual packet data plus metadata (frequency, datarate, snr, rssi, tmst, and GPS time/location). All this data must be available to validate the signature. This PR proposes adding an optional `secure_packet_v1` message type to the `lora_beacon_report_req_v1`

### DIFF
--- a/src/service/poc_lora.proto
+++ b/src/service/poc_lora.proto
@@ -26,6 +26,56 @@ enum invalid_reason {
   invalid_frequency = 13;
 }
 
+message secure_packet_v1 {
+  // frequency in Hz
+  uint64 freq = 1;
+
+  data_rate datarate = 2;
+
+  // Signal to Noise radio (0.01dB)
+  int32 snr = 3;
+
+  // Received Signal Strength Indicator (1 dBm)
+  int32 rssi = 4;
+
+  // Internal timestamp of "RX finished" event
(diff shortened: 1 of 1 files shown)


src/service/poc_lora.proto
```diff
@@ -26,6 +26,56 @@ enum invalid_reason {
...
+  int32 height = 3;
+
+  // Horizontal accuracy estimate (mm)
+  uint32 hacc = 4;
+
+  // Vertical accuracy estimate (mm)
+  optional uint32 vacc = 5;
+}
+
+// Duration since the GPS epoc (0h 6-Jan-1980)
```
abhay, line 73: Might be worthwhile to use the proto timestamp type here

lthiery: For reference: 
https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/timestamp.proto#L133

abhay, line 73: removed comments about the well known type... better to look at this proposed guideline https://github.com/helium/proto/pull/208

devilish66694 (approved):

PR #196, opened 2022-10-07
Add rewards report to mobile services proto

maplant:
Addresses https://github.com/helium/oracles/issues/117 (hopefully)

### DIFF
--- a/src/service/poc_mobile.proto
+++ b/src/service/poc_mobile.proto
@@ -163,11 +163,19 @@ message rewards_report {
 }
 
 message hotspot_reward_report {
+  /// Public key of the hotspot
   bytes pub_key = 1;
-  uint64 reward = 2;
+  /// Public key of the owner of the hotspot
+  bytes owner = 2;
+  /// Reward amount in bones
+  uint64 reward = 3;
 }
 
 message radio_reward_report {
+  /// Public key of the radio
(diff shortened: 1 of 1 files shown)


src/service/poc_mobile.proto
```diff
@@ -156,3 +156,18 @@ enum cell_type {
...
 }
+
+message rewards_report {
+  repeated hotspot_reward_report hotspot_rewards = 1;
+  repeated radio_reward_report radio_rewards = 2;
+}
+
+message hotspot_reward_report {
+  bytes pub_key = 1;
+  uint64 reward = 2;
```
madninja, line 167: Add what this is? Mobile Bones presumably

madninja: I think you’re missing the owner account that’s being rewarded. The hotspot itself isn’t given tokens

madninja (changes requested):
I'm not sure this is right.. 

1. rewards report is per radio, hotspot and owner.. one or more radios can be associated with a hotspot so expect multiple of the same hotspot key but with different radio ids. 

2. a radio is identified by it's cbsd_id (String) not a public key

3. Just to be clear,  this reward_report is what mobile_rewards emits. The verifier should emit reward shares for for each of these with some magic around what reward period they belong to?

PR #199, opened 2022-10-11
Make RSSI a sint32

macpie:


src/service/packet_router.proto
```diff
@@ -8,7 +8,7 @@ import "data_rate.proto";
 message packet_router_packet_up_v1 {
   bytes payload = 1;
   uint64 timestamp = 2;
-  uint32 rssi = 3;
+  sint32 rssi = 3;
```
dpezely, line 11: Add comment indicating units:
```suggestion
  // LoRa signal strength in dB.
  sint32 rssi = 3;
```

src/service/packet_router.proto
```diff
@@ -24,7 +24,7 @@ message packet_router_packet_report_v1 {
   uint32 oui = 2;
   uint64 net_id = 3;
   // LoRa signal strength
```
dpezely, line 26: Add units:
```suggestion
  // LoRa signal strength in dB.
```

dpezely (approved):
Add comments that include units, per various discussions over this past summer.

src/service/packet_router.proto
madninja: should this be LoRa Signal Strength in `ddbi`?

michaeldjeffrey: semtech protocol marks it as `dBm`(?)
https://github.com/Lora-net/packet_forwarder/blob/master/PROTOCOL.TXT#L144

PR #208, opened 2022-10-13
Guide for contributors

dpezely:
When and why to use specific types and units for various values.

The initial commit is just a rough draft.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,3 +2,18 @@
 ![rust workflow](https://github.com/helium/proto/actions/workflows/rust.yml/badge.svg)
 ![node.js workflow](https://github.com/helium/proto/actions/workflows/node.yml/badge.svg)
 ![erlang workflow](https://github.com/helium/proto/actions/workflows/erlang.yml/badge.svg)
+
+## Contributing
+
+- Avoid `float` in Protobufs because representations differ across hardware architectures
+  + There are many floating point representations from IEEE, plus others
+  + i.e., some range of interior digits are random per float spec
+- Frequency should use `uint32`
+  + Locally, define accessor functions to return float after appropriate division
+- rssi or signal is always negative, thus use `sint32`
+- snr is signal-to-noise ratio and should be `uint32`
+- Fetch and share time in nanos, then truncate to appropriate granularity as needed
+  + e.g., get from OS in nanos
+- Reject any PR unless units are documented inline within Protobuf definition
+- Append units suffix on field names when deviating from the above descriptions
+- When exceptions to the above occur, please explain *why* within comments
--- a/README.md
+++ b/README.md
@@ -9,11 +9,10 @@
   + There are many floating point representations from IEEE, plus others
   + i.e., some range of interior digits are random per float spec
 - Frequency should use `uint32` and should be in Hz
-  + Locally, define accessor functions to return float after appropriate division
 - rssi or signal is always negative, thus use `sint32` and is in deci-dbm (aka `ddbm`) which is `dbm * 10`
 - snr is signal-to-noise ratio and should be `uint32`
 - Fetch and share time in nanos, then truncate to appropriate granularity as needed
   + e.g., get from OS in nanos
 - Reject any PR unless units are documented inline within Protobuf definition
-- Append units suffix on field names when deviating from the above descriptions
+- Document units of fields
 - When exceptions to the above occur, please explain *why* within comments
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,3 +2,18 @@
...
+  + There are many floating point representations from IEEE, plus others
+  + i.e., some range of interior digits are random per float spec
+- Frequency should use `uint32`
+  + Locally, define accessor functions to return float after appropriate division
+- rssi or signal is always negative, thus use `sint32`
+- snr is signal-to-noise ratio and should be `uint32`
+- Fetch and share time in nanos, then truncate to appropriate granularity as needed
+  + e.g., get from OS in nanos
+- Reject any PR unless units are documented inline within Protobuf definition
+- Append units suffix on field names when deviating from the above descriptions
```
madninja, line 18: no.. I'm still against this since we'd have to so many of them to have it not be some random handful.. 

```
Document units of fields
```

README.md
```diff
@@ -2,3 +2,18 @@
...
 ![erlang workflow](https://github.com/helium/proto/actions/workflows/erlang.yml/badge.svg)
+
+## Contributing
+
+- Avoid `float` in Protobufs because representations differ across hardware architectures
+  + There are many floating point representations from IEEE, plus others
+  + i.e., some range of interior digits are random per float spec
+- Frequency should use `uint32`
+  + Locally, define accessor functions to return float after appropriate division
+- rssi or signal is always negative, thus use `sint32`
```
madninja, line 13:
```suggestion
- rssi or signal is always negative, thus use `sint32` and is in deci-dbm (aka `ddbm`) which is `dbm * 10`
```

README.md
```diff
@@ -2,3 +2,18 @@
...
+- snr is signal-to-noise ratio and should be `uint32`
```
madninja, line 14: We need to add level of precision for this ratio? or is it always a whole multiplier

README.md
```diff
@@ -2,3 +2,18 @@
...
+- Fetch and share time in nanos, then truncate to appropriate granularity as needed
```
abhay, line 15: `truncate to appropriate granularity`: does this mean 

I've had success using the ["well known type"](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/timestamp.proto#L133) in other projects and letting clients figure out what to do with the extra precision.

madninja: we don't share time in nanos in all protos.. some are millis, nanos or secs depending on whether sharing something finer grained makes sense (or is available). The doc string for each timestamp indicates what it is and the rust `file_store::MsgTimestamp` trait can be used to ensure you decode/encode the timestamp from the right units to `DateTime<Utc>`.

We're sort of stuck with this pattern for now since we already have active use of it so introducing partial switchovers doesn't seem right currently

abhay: staying with this pattern makes sense given this additional context.

README.md
dpezely: Changed in commit a2fe4d5

README.md
```diff
@@ -2,3 +2,18 @@
...
 ![node.js workflow](https://github.com/helium/proto/actions/workflows/node.yml/badge.svg)
 ![erlang workflow](https://github.com/helium/proto/actions/workflows/erlang.yml/badge.svg)
+
+## Contributing
+
+- Avoid `float` in Protobufs because representations differ across hardware architectures
+  + There are many floating point representations from IEEE, plus others
+  + i.e., some range of interior digits are random per float spec
+- Frequency should use `uint32`
+  + Locally, define accessor functions to return float after appropriate division
```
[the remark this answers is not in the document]
dpezely: Removed in commit a2fe4d5

README.md
dinocore1: I think signal-to-noise ratio should be sint32 and in units of deci-db. I believe LoRa radios can actually receive below the noise floor.

README.md
dinocore1: Not sure if it applicable here, but I wanted to add that some LoRa Concentrators (w/ SX1303) can record fine-timestamp of received packets. The fine-timestamp has nanosecond precision and could be someday useful for calculating Time-Difference-of-Arrival. To perform the TDoA calculation, we need to preserve the nanosecond precision.

madninja: That's correct and why witness and beacon report timestamps are in nanos

README.md
madninja: doesn't matter for the type instruction here though

madninja (approved):

PR #209, opened 2022-10-18
push gateway not found lookup result to the app level

jeffgrunewald:
pre-requisite to https://github.com/helium/blockchain-node/pull/181, allows the response from the blockchain-node to distinguish between a result with location and region information and when the ledger is unable to verify the gateway by the requested pubkey

### DIFF
--- a/src/service/follower.proto
+++ b/src/service/follower.proto
@@ -7,6 +7,12 @@ import "blockchain_txn.proto";
 import "gateway_staking_mode.proto";
 import "region.proto";
 
+message gateway_not_found { bytes address = 1; }
+
+message follower_error {
+  oneof type { gateway_not_found not_found = 1; }
+}
+
 message follower_txn_stream_req_v1 {
   uint64 height = 1;
   bytes txn_hash = 2;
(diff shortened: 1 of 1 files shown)


src/service/follower.proto
```diff
@@ -46,6 +38,23 @@ message follower_gateway_resp_v1 {
   region region = 7;
 }
 
+message gateway_not_found {}
```
andymck, line 41: might be better to include a generic error response which can cater to any future error handling ( including any additional APIs which get added over time)...something like:

enum follower_error {
  gateway_not_found = 0;
}

message follower_error_resp {
  follower_error error = 1;  
  bytes details = 2;
}

In the case of a gateway not found, the error would equal the gateway_not_found enum value, whilst details would be the pub key specified in the request

jeffgrunewald: done [54c1367](https://github.com/helium/proto/pull/209/commits/54c136709e9570979757fe2d1baf972bf98aa15a)

src/service/follower.proto
```diff
@@ -7,6 +7,13 @@ import "blockchain_txn.proto";
 import "gateway_staking_mode.proto";
 import "region.proto";
 
+enum follower_error { gateway_not_found = 0; }
+
+message follower_error_resp {
```
madninja, line 12: Why is this here? What would bytes contain?

jeffgrunewald: See @andymck ’s previous comment; contextual but in this case would be used to send back the pubkey that was missing in case the caller was making multiple calls and needed to differentiate which gateway wasn’t found

madninja: Yeah, I really dislike the "blob of contextual data" approach here vs errors with actual details in the message type itself.

jeffgrunewald: I can remove it but I’m disinclined to gold-plate this since it’s such a short-lived solution. Lemme play around briefly with a more structured not_found error than just an enum

jeffgrunewald: adjusted here to make the error type more distinct: [0896e48](https://github.com/helium/proto/pull/209/commits/0896e488d62c86a2978b8546b268d5aa4608388d); lemme know what you think

madninja (approved):

PR #212, opened 2022-10-20
Add speedtest_avg and speedtest_avg_validity

maplant:
Also, removes speed_share which is not very useful

### DIFF
--- a/src/service/poc_mobile.proto
+++ b/src/service/poc_mobile.proto
@@ -95,6 +95,7 @@ message speedtest_avg {
   uint32 latency_avg_ms = 4;
   speedtest_avg_validity validity = 5;
   repeated speedtest speedtests = 6;
+  uint64 timestamp = 7;
 }
 
 message speedtest {
--- a/src/service/poc_mobile.proto
+++ b/src/service/poc_mobile.proto
@@ -77,26 +77,6 @@ message share {
 
 message shares { repeated share shares = 1; }
 
-message speed_share {
-  bytes pub_key = 1;
-  /// bytes/sec
-  uint64 upload_speed_bps = 2;
-  /// bytes/sec
-  uint64 download_speed_bps = 3;
-  uint32 latency_ms = 4;
-  uint64 timestamp = 5;
-  share_validity validity = 6;
-}
-
-message speed_share_list {
-  bytes gw_pub_key = 1;
-  repeated speed_share speed_shares = 2;
-}
-
-message speed_shares { repeated speed_share_list speed_shares = 1; }
-
-message invalid_speed_shares { repeated speed_share speed_shares = 1; }
-
 enum share_validity {
   share_validity_valid = 0;
   share_validity_decode_failed = 1;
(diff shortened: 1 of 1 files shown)


bbalser (approved):
LGTM

@meowshka

src/service/poc_mobile.proto
```diff
@@ -106,22 +86,31 @@ enum share_validity {
   share_validity_not_operational = 5;
 }
 
-message moving_avg {
+message speedtest_avg {
   bytes pub_key = 1;
-  /// Refers to the number of speed_shares used to calculate the average
-  uint64 window_size = 2;
```
meowshka, line 112: Any particular reason why window_size was removed? We might use it in Explorer to tell user on how many datapoints their average is calculated.

src/service/poc_mobile.proto
```diff
@@ -106,22 +86,31 @@ enum share_validity {
   share_validity_not_operational = 5;
 }
 
-message moving_avg {
+message speedtest_avg {
```
meowshka, line 89: Can we add a timestamp of when the Speed Test average calculation happened?

src/service/poc_mobile.proto
```diff
@@ -106,22 +86,31 @@ enum share_validity {
...
+message speedtest {
+  uint64 upload_speed_bps = 1;
+  uint64 download_speed_bps = 2;
+  uint32 latency_ms = 3;
+  uint64 timestamp = 4;
 }
 
-message speed_share_moving_avgs { repeated moving_avg moving_avgs = 1; }
+enum speedtest_avg_validity {
+  speedtest_avg_validity_valid = 0;
```
meowshka, line 108: Is this value derived from either of these being true speedtest_avg_validity_slow_download_speed, speedtest_avg_validity_slow_upload_speed, speedtest_avg_validity_high_latency?

src/service/poc_mobile.proto
bbalser: the window size is the number of speedtests attached to the average, so we removed this since it is duplicated information.

meowshka: Understood, thanks!

src/service/poc_mobile.proto
maplant: Sure

src/service/poc_mobile.proto
maplant: That value would be if none of the others are true

src/service/poc_mobile.proto
```diff
@@ -106,22 +86,32 @@ enum share_validity {
...
-  bool is_valid = 4;
-  repeated speed_share speed_shares = 5;
+  /// bytes/sec
+  uint64 upload_speed_avg_bps = 2;
+  /// bytes/sec
+  uint64 download_speed_avg_bps = 3;
+  uint32 latency_avg_ms = 4;
+  speedtest_avg_validity validity = 5;
+  repeated speedtest speedtests = 6;
+  uint64 timestamp = 7;
```
bbalser, line 98: please add a comment in what units this timestamp is in, I assume millis

bbalser (approved):

PR #215, opened 2022-10-24
add CPP build workflow

malcolmmadsheep:


madninja:
The issue we were debugging with this workflow actually did surface with the other languages as well.. We appeared to be on the wrong branch to catch it. 

I'm fine merging this but it would need a rebase and fix up the changes that have happened on the main branch since this PR was created. Would you be a able to do that @malcolmmadsheep ?

malcolmmadsheep:
@madninja hi, sure, no problem

madninja (approved):

PR #218, opened 2022-10-26
Additional configuration for HPR http roaming protocol

ruttenberg:
This adds two fields and an enum definition.
It is only used by Helium Packet Router.


dpezely (approved):

src/service/config.proto
```diff
@@ -23,7 +24,11 @@ message protocol_gwmp_mapping_v1 {
 
 message protocol_gwmp_v1 { repeated protocol_gwmp_mapping_v1 mapping = 1; }
 
-message protocol_http_roaming_v1 {}
+message protocol_http_roaming_v1 {
+  flow_type flow_type = 1;
+  uint32 dedupe_timeout = 2;
```
madninja, line 29: declare in what unit in a comment

src/service/config.proto
```diff
@@ -23,7 +24,11 @@ message protocol_gwmp_mapping_v1 {
...
+  bytes path = 3;
```
madninja, line 30: declare what this is? if it's an actual http endpoint/path why is this not a string?

ruttenberg: I was unaware of "string." I will take a look.

src/service/config.proto
ruttenberg: Will do.

src/service/config.proto
ruttenberg: I have changed it to "string".

src/service/config.proto
ruttenberg: Done.

madninja (approved):

PR #220, opened 2022-10-27
Mapper Protos: move out from services, rename, add some fields to attach_response.

lthiery:
Some fields were ommitted from attach_response.


src/mapper.proto
```diff
@@ -0,0 +1,103 @@
+syntax = "proto3";
+package helium.mapper;
+
+import "mapper_attach.proto";
```
madninja, line 4: These no longer needed ?

madninja (commented):
Looks like some of the imports are wrong but overall this looks great

madninja (approved):

PR #222, opened 2022-10-28
Change subnetwork reward to give out reward percentage

maplant:
I'm fairly certain that a double float is suitable enough to describe reward period percentages

### DIFF
--- a/src/decimal.proto
+++ b/src/decimal.proto
@@ -1,0 +1,9 @@
+syntax = "proto3";
+package helium;
+
+/// rust_decimal serializes to an array of 16 bytes. We encode that as two uint64s
+/// taking the bytes in big-endian (arbitrarily chosen).
+message decimal {
+  uint64 hi = 1;
+  uint64 lo = 2;
+}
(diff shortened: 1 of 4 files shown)


madninja (commented):
This is not a safe change:

1. This is being used by the existing L1 chain. Changing the proto will break chain behavior 
2. doubles/floats in protos are against the proto rules, given how they're not accurate across architectures which will break validation on the existing chain, or may behave in very unexecpected ways in the new world

maplant:
> This is not a safe change:
> 
>     1. This is being used by the existing L1 chain. Changing the proto will break chain behavior
> 
>     2. doubles/floats in protos are against the proto rules, given how they're not accurate across architectures which will break validation on the existing chain, or may behave in very unexecpected ways in the new world

This should be only additions, subnetwork rewards should not be modified. 

I'll change this to use decimals

src/decimal.proto
```diff
@@ -0,0 +1,9 @@
+syntax = "proto3";
+package helium;
+
+/// rust_decimal serializes to an array of 16 bytes. We encode that as two
+/// uint64s taking the bytes in big-endian (arbitrarily chosen).
+message decimal {
```
madninja, line 6: This is certainly not convention today. Let’s talk before you push this in elsewhere

PR #230, opened 2022-11-08
convert wire-format of scaling factor to string

jeffgrunewald:
within the application code, to avoid imprecise conversions for rewards we use the Decimal implementation defined in `rust_decimal`; the string format of the type when serialized over the wire is likely to lead to more precise interpretations of the intended value across different systems/architectures than converting to and from floating point types.


JayKickliter (changes requested):
Thanks, I think this is a better idea than using floats, even though strings are little weird. At a minimum, please add a comment above the fields explaining what this string represents (numerically), so future us aren't scratching our heads

JayKickliter (approved):

PR #231, opened 2022-11-09
convert scaling factors to integers

jeffgrunewald:
to avoid precision conflicts across architectures, represent these scaling factor percentages as integers; in practice they are a 2-decimal precision percentage multiplier.

### DIFF
--- a/src/scaling_factors.proto
+++ b/src/scaling_factors.proto
@@ -2,4 +2,6 @@ syntax = "proto3";
 
 package helium;
 
-message scaling_factors { map<string, double> hex_to_sf = 1; }
+// map of string h3 res 12 hex address to reward scaling factor
+// as an integer representation of the 2-point precision decimal percent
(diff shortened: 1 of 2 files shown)


src/scaling_factors.proto
```diff
@@ -2,4 +2,6 @@ syntax = "proto3";
 
 package helium;
 
-message scaling_factors { map<string, double> hex_to_sf = 1; }
+// map of string h3 res 12 hex address to reward scaling factor
+// as an integer representation of the 2-point precision decimal percent
```
madninja, line 6: why not use a 4-point precision decimal float vs a percentage? I know they're equivalent but a single 4 point precision decimal is easier to ser/deser?

jeffgrunewald: adjusted and clarified in the comment [eeba244](https://github.com/helium/proto/pull/231/commits/eeba244d8817a7eeaa1997441872269b22488b88)

madninja (approved):

PR #234, opened 2022-11-14
add reward unit to beacon and witness reports

jeffgrunewald:
necessary for moving the calculation of the reward unit (a multiplier of iot reward shares) currently encapsulated within the Iot Injector but being moved to the IoT Verifier. Rather than multiplying the `hex_scale` and `reward_unit` values within the verifier and sending the values as a single field in the outgoing valid reports for ease of upgrading the existing proto messages as well as additional transparency they are sent as separate fields.

### DIFF
--- a/src/service/poc_lora.proto
+++ b/src/service/poc_lora.proto
@@ -90,6 +90,9 @@ message lora_valid_beacon_report_v1 {
   // ex: 5015 == 0.5015
   uint32 hex_scale = 3;
   lora_beacon_report_req_v1 report = 4;
+  // integer representation of a 4-point precision decimal multiplier
+  // based on the number of witnesses to a poc event
+  uint32 reward_unit = 5;
 }
 
 // tagged valid witness report produced by the verifier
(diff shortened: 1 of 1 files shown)


src/service/poc_lora.proto
```diff
@@ -90,6 +90,9 @@ message lora_valid_beacon_report_v1 {
   // ex: 5015 == 0.5015
   uint32 hex_scale = 3;
   lora_beacon_report_req_v1 report = 4;
+  // integer representation of a 4-point precision decimal multiplier
+  // basdd on the number of witnesses to a poc event
```
andymck, line 94: nit: typo

andymck (approved):

src/service/poc_lora.proto
vihu: Does this have to be 4-point? This `reward_unit` is the same as https://github.com/helium/proto/blob/master/src/blockchain_poc_core_v1.proto#L26 correct?

jeffgrunewald: I’m looking to confirm that or not; arguably the erlang equivalent of the hex_scale values is only 2 decimals since it’s multiplying and then dividing by 100 but we opted to be more precise. I don’t know that it’s necessary or not in this case but i at least don’t think it _inappropriate_ to be more precise. We’re talking about scaling multipliers going from 0.0152 to 0.02 for example if we only handle 2 points precision on the hex_scale

dpezely (approved):

madninja (approved):

PR #235, opened 2022-11-15
Add inventory message

maplant:


maplant:
> I think this should be consistent across both mobile and iot network reward period strategies so it should probably be in its own file in the `src/` like `src/rewards_manifest.proto` or something. We should workshop the name b/c at least for now the iot `written_files` picked up by the injector will be poc reports and only later will it be reward shares reports like the mobile indexer will read but something like that

I think that this naming is generalized enough to be used anywhere

madninja (commented):
Bit too generic of a name right? 

I see it can be used for multiple purposes. But it's basically a file_inventory? or a file_range? 

And it's certainly not a start_end "epoch". Those are timestamps so just call them `start/end_timestamp`?

maplant:
> I was thinking `rewardable_inventory.proto` or `reward_manifest.proto` that defines an inventory message so it indicates the context of the file and isn’t easily confused with something for managing device inventories or something. I think `start_/end_timestamp` also makes sense as we move away from chain-based time periods

I renamed this struct reward_manifest and renamed the start and end values to _timestamp

jeffgrunewald (approved):

PR #246, opened 2022-12-15
OUI Balance Stream and Verified Packet 

maplant:


src/packet.proto
```diff
@@ -36,3 +36,11 @@ message packet {
   routing_information routing = 9;
   window rx2_window = 10;
 }
+
+message verified_packet {
```
madninja, line 40: Unless this is shared across services outside of oracles, which I don't think it is, move this to the packet verifier service?

src/service/follower.proto
```diff
@@ -89,4 +100,6 @@ service follower {
   rpc subnetwork_last_reward_height(
       follower_subnetwork_last_reward_height_req_v1)
       returns (follower_subnetwork_last_reward_height_resp_v1);
+  rpc oui_balance_stream(follower_oui_balance_update_req_v1)
```
madninja, line 103: 1. What's the bootstrap here? How do you get the initial balance? Streamed at first connect? for all router keys and their oui?
2. What step do we need the dependency in again vs getting it from the SOL chain?

maplant: 1. The first message from the follower gives you the balance of all of the oui
2. It's for checking the balance to determine if a packet is valid. It would get replace outright with checking the balance on the SOL chain

madninja: > * The first message from the follower gives you the balance of all of the oui

Gotcha, ok

> * It's for checking the balance to determine if a packet is valid. It would get replace outright with checking the balance on the SOL chain

Good for now. Just put it in a function we can easily replaces  with a solana-sdk one

andymck:
@maplant after a discussion today, i have simplified the proto for tracking balances.  API is now a unary request where the client can specify a list of accounts to derive balances for.
NOTE: this is still ongoing discussion as to whether we will actually require this

andymck:
@maplant a draft of the node API implementation to query balances is available here: https://github.com/helium/blockchain-node/pull/192

PR #247, opened 2022-12-16
add devaddr_ranges field to org_v1

michaeldjeffrey:
A "Helium" org would have 1-n ranges within the Helium NetID. "Roaming" orgs would have a single range that spans their entire NetID.

Any Routes created or updated will have their devaddr ranges checked against the orgs list of devaddr range restrictions.

In thinking about calling this field "restrictions", there could be more fields for an org that come along and "restrict" their ability to do various things. In other messages, devaddr_ranges are named the same, and the intent can be easily inferred through the context of the message.

### DIFF
--- a/src/service/iot_config.proto
+++ b/src/service/iot_config.proto
@@ -8,12 +8,29 @@ import "region.proto";
 // Message Definitions
 // ------------------------------------------------------------------
 
+// == Field Notes ==
+//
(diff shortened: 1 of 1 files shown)


jeffgrunewald:
Are you saying “in other messages `devaddr_ranges` are named “devaddr_ranges` so context would supply the connection between the two different names?

macpie (approved):

PR #250, opened 2022-12-19
Define custom regions

mikev:
Problem Statement:
Often a country or several countries need a custom region, which defines slightly different parameters.  For example in Malaysia any frequencies in 924 Mhz are not legal.

Solution:
So we defined a new region AS923_1B which did not use frequency within 924 Mhz.  Now Kenya needs this same custom region, because some of our EU868 frequencies are not legal.  This will be a recurring pattern.  The PR, review, commit process is painful so lets just do this once across all our repos and define EU868_A to _F and AS923_1A .. 1F

Only AS923 and EU868 will require custom regions.

Libraries (Require a review 1st)
https://github.com/helium/proto/pull/250
https://github.com/helium/blockchain-core/pull/1502
https://github.com/helium/erlang-lorawan/pull/27

Projects
https://github.com/helium/miner/pull/1873
https://github.com/helium/gateway-rs/search?q=as923 [gateway-rs no longer requires enum]
https://github.com/helium/sx1302_hal/pull/36
https://github.com/helium/docs/pull/1048/files
https://github.com/helium/helium-config-service/pull/27 (edited)


ke6jjj (approved):

madninja:
you ready for this to be reviewed/merged @mikev ?

ke6jjj:
We've got about 7 pulls that should be reviewed together, first, @madninja

PR #252, opened 2022-12-20
add missing region params message

jeffgrunewald:
This update to the lib.rs and build.rs are needed to allow the generated code to properly import the struct type of the region params newly added to the iot_config service and to do deserialization of the type.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -9,12 +9,13 @@ pub use prost::{DecodeError, EncodeError, Message};
 #[cfg(feature = "services")]
 pub mod services {
     use crate::{
-        BlockchainTokenTypeV1, BlockchainTxn, DataRate, GatewayStakingMode, MapperAttach, Region,
-        RoutingAddress,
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -9,12 +9,13 @@ pub use prost::{DecodeError, EncodeError, Message};
...
     use crate::{
-        BlockchainTokenTypeV1, BlockchainTxn, DataRate, GatewayStakingMode, MapperAttach, Region,
-        RoutingAddress,
+        BlockchainRegionParamsV1, BlockchainTokenTypeV1, BlockchainTxn, DataRate,
+        GatewayStakingMode, MapperAttach, Region, RoutingAddress,
     };
 
     pub mod iot_config {
         include!(concat!(env!("OUT_DIR"), "/helium.iot_config.rs"));
+        pub use gateway_server::{Gateway, GatewayServer};
```
madninja, line 18: What is GatewayServer here? it's not something any of the iot oracles run right?

madninja: ah ok yeah, this is fine to keep consistent

madninja (approved):

PR #256, opened 2023-01-05
support generic error as invalid reason

andymck:
Adds general error to be used as the invalid reason for a witness or beacon should verification hit an unexpected error and no specific verification fail reason can be determined

### DIFF
--- a/src/service/poc_lora.proto
+++ b/src/service/poc_lora.proto
@@ -27,6 +27,7 @@ enum invalid_reason {
   self_witness = 14;
   stale = 15;
   scaling_factor_not_found = 16;
+  unknown_error = 17;
 }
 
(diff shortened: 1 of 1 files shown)


src/service/poc_lora.proto
```diff
@@ -27,6 +27,7 @@ enum invalid_reason {
   self_witness = 14;
   stale = 15;
   scaling_factor_not_found = 16;
+  generic_error = 17;
```
madninja, line 30: Rename to "unknown". Generic makes it sound like it has a meaningful use and we're really trying to indicate it's an error of unknown origin.

andymck: done, ocd satisfied

madninja (approved):

PR #257, opened 2023-01-05
annotate deprecate nonce field checks

jeffgrunewald:
For the time being, we have decided to deprecate checking the route and org nonce is incremented on updates in favor of exclusively validating the timestamp supplied in the signed request is within the configured ttl window _and_ is ordered monotonically increasing from the previous update to the record.

Annotate the nonce field in the proto repo to ensure that consumers of the proto repo are aware the field is no longer used.

### DIFF
--- a/src/service/iot_config.proto
+++ b/src/service/iot_config.proto
@@ -47,7 +47,8 @@ message org_v1 {
   bytes payer = 3;
   // List of keys allowed some specific actions, see services.
   repeated bytes delegate_keys = 4;
-  uint64 nonce = 5;
+  // Incremental int for updates (DEPRECATED)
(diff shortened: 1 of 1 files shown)


macpie (approved):

madninja (changes requested):
* comments go above the field
* Deprecated is supported as a proto attribute. See https://github.com/helium/proto/blob/master/src/region.proto#L10

jeffgrunewald:
> * comments go above the field
> * Deprecated is supported as a proto attribute. See https://github.com/helium/proto/blob/master/src/region.proto#L10

excellent; much better. done

madninja (approved):

PR #258, opened 2023-01-06
Remove timestamps from data_transfer_session_req_v1

bbalser:


madninja:
Why remove the timestamp? It removes the ability to locate the data usage on a timeline and could make for replay attacks

bbalser:
The way I understand it, there is no start or end timestamp associated with the request that will be sent to the ingestor.  The start and end timestamp they have is for the telco session that does not line up with the requests we will receive.  When we talked about this, you agreed that the received timestamp would be enough for us to associate the data transfer with a rewardable period.  Maybe I misunderstood what we talked about.

madninja:
> The way I understand it, there is no start or end timestamp associated with the request that will be sent to the ingestor. The start and end timestamp they have is for the telco session that does not line up with the requests we will receive. When we talked about this, you agreed that the received timestamp would be enough for us to associate the data transfer with a rewardable period. Maybe I misunderstood what we talked about.

The concern I have is that I can submit the same signed request again and we won’t know (ie replay)

bbalser:
ok, I will just add a timestamp back in and they can just set it to the time they send us the request.

madninja (approved):

PR #259, opened 2023-01-07
derive deser for region (and others)

jeffgrunewald:
implement deserialize alongside serialize for the core proto messages used throughout various projects to simplify and consolidate definitions of the commonly used protos into the proto repo and avoid having to re-implement them or else wrap them in new types to allow defining common traits like (de)serialize

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -32,7 +32,10 @@ fn main() -> Result<()> {
     tonic_build::configure()
         .build_server(true)
         .build_client(true)
-        .type_attribute(".", "#[derive(serde_derive::Serialize)]")
+        .type_attribute(
+            ".",
+            "#[derive(serde_derive::Serialize, serde_derive::Deserialize)]",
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -32,7 +32,10 @@ fn main() -> Result<()> {
     tonic_build::configure()
         .build_server(true)
         .build_client(true)
-        .type_attribute(".", "#[derive(serde_derive::Serialize)]")
+        .type_attribute(
+            ".",
+            "#[derive(serde_derive::Serialize, serde_derive::Deserialize)]",
+        )
         .type_attribute(".helium.config", "#[derive(serde_derive::Deserialize)]")
```
madninja, line 39: Remove the `.helium.config` attribute since the `.` pattern match in this PR adds it to all protos

madninja (approved):

PR #263, opened 2023-01-13
Scale Routes

macpie:
* Routes dont include euis and devaddr range anymore
* Stream EUIs and ranges update for scale

### DIFF
--- a/src/service/iot_config.proto
+++ b/src/service/iot_config.proto
@@ -22,14 +22,15 @@ import "region.proto";
 //
 // == DevAddrs ==
 //
-// - `devaddr_range` are inclusive on both sides, `start_addr` and `end_addr`.
+// - `devaddr_range` and `devaddr_constraints` are inclusive on both sides,
+// - `start_addr` and `end_addr`.
 //
-// - `org_res_v1.devaddr_range` provides the ranges of DevAddrs available to
+// - `org_res_v1.devaddr_constraints` provides the ranges of DevAddrs available to
 //    any `route_v1` under an Org.
 //
-// - `route_v1.devaddr_range` provides the ranges of DevAddrs that should go to
+// - `devaddr_range_v1` provides the ranges of DevAddrs that should go to
 //    this specific Route. This ranges must always fall within the
-//    `devaddr_range` of the owning org.
+//    `devaddr_constraints` of the owning org.
 
 enum action_v1 {
   add = 0;
@@ -52,6 +53,11 @@ message devaddr_range_v1 {
   string route_id = 1;
   uint32 start_addr = 2;
   uint32 end_addr = 3;
+}
+
+message devaddr_constraint_v1 {
+  uint32 start_addr = 1;
+  uint32 end_addr = 2;
 }
 
 // Device App EUI and Dev EUI
@@ -148,7 +154,7 @@ message org_create_roamer_req_v1 {
 message org_res_v1 {
   org_v1 org = 1;
   uint64 net_id = 2;
-  repeated devaddr_range_v1 devaddr_ranges = 3;
+  repeated devaddr_constraint_v1 devaddr_constraints = 3;
 }
 
 message org_disable_req_v1 {
--- a/src/service/iot_config.proto
+++ b/src/service/iot_config.proto
@@ -402,7 +402,7 @@ service route {
 
   // Get DevAddr Ranges for a Route (auth delegate_keys/owner/admin)
   rpc get_devaddr_ranges(route_get_devaddr_ranges_req_v1)
-      returns (stream eui_pair_v1);
+      returns (stream devaddr_range_v1);
   // Update (single add or remove) DevAddr Ranges for a Route (auth
   // delegate_keys/owner/admin)
   rpc update_devaddr_ranges(stream route_update_devaddr_ranges_req_v1)
--- a/src/service/iot_config.proto
+++ b/src/service/iot_config.proto
@@ -275,7 +275,7 @@ message route_stream_res_v1 {
   action_v1 action = 1;
   oneof data {
     route_v1 route = 2;
-    eui_pair_v1 euis = 3;
+    eui_pair_v1 eui_pair = 3;
     devaddr_range_v1 devaddr_range = 4;
   }
 }
--- a/src/service/iot_config.proto
+++ b/src/service/iot_config.proto
@@ -415,7 +415,7 @@ service route {
   rpc stream(route_stream_req_v1) returns (stream route_stream_res_v1);
 }
 
-// This service will be update soon (APIs not accurate as of 2023/01/13)
+// This service will be updated soon, APIs are not accurate (2023/01/13)
 service session_key_filter {
   // List Filters for an Org (auth delegate_keys/owner/admin)
   rpc list(session_key_filter_list_req_v1)
--- a/src/service/iot_config.proto
+++ b/src/service/iot_config.proto
@@ -56,7 +56,7 @@ message devaddr_range_v1 {
 }
 
 // Device App EUI and Dev EUI
-message euis_v1 {
+message eui_pair_v1 {
   string route_id = 1;
   uint64 app_eui = 2;
   uint64 dev_eui = 3;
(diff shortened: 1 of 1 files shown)


src/service/iot_config.proto
```diff
@@ -40,27 +40,26 @@ enum action_v1 {
...
+  string route_id = 1;
+  uint32 start_addr = 2;
+  uint32 end_addr = 3;
 }
 
 // Device App EUI and Dev EUI
-message eui_v1 {
-  uint64 app_eui = 1;
-  uint64 dev_eui = 2;
+message euis_v1 {
```
jeffgrunewald, line 59:
```suggestion
message eui_v1 {
```
i like keeping this singular so you don't see it referenced in the context of the code and mistake it for a message/field comprising multiple eui pairs. could rename `eui_pair_v1` if you want to reflect the fact it contains two eui fields

jeffgrunewald (commented):
this looks good; makes a lot of sense based on interactions with the volume of changes you expect for these associated types;
my only question is the semantics of the singular vs. plural req types. i _think_ i'd lean toward what you've done here since most of these request messages being made in the context of a stream, but the update request is like `route_update_euis_req_v1` when each message is only containing an update for a single eui pair so it's tricky

src/service/iot_config.proto
macpie: sounds good to me

src/service/iot_config.proto
```diff
@@ -210,50 +205,68 @@ message route_delete_req_v1 {
...
 
-message route_euis_req_v1 {
-  string id = 1;
-  route_euis_action_v1 action = 2;
-  repeated eui_v1 euis = 3;
+message route_update_euis_req_v1 {
+  enum route_update_euis_action_v1 {
+    add_euis = 0;
+    remove_euis = 1;
+  }
```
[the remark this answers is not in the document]
macpie: yep I can do that

src/service/iot_config.proto
```diff
@@ -378,19 +386,36 @@ service route {
...
+  // Update (single add or remove) EUIs for a Route (auth
   // delegate_keys/owner/admin)
-  rpc devaddrs(route_devaddrs_req_v1) returns (route_devaddrs_res_v1);
+  rpc update_euis(stream route_update_euis_req_v1) returns (route_euis_res_v1);
+  // Delete all EUIs for a Route (auth delegate_keys/owner/admin)
+  rpc delete_euis(route_delete_euis_req_v1) returns (route_euis_res_v1);
+
+  // Get DevAddr Ranges for a Route (auth delegate_keys/owner/admin)
+  rpc get_devaddr_ranges(route_get_devaddr_ranges_req_v1)
+      returns (stream eui_pair_v1);
```
jeffgrunewald, line 405:
```suggestion
      returns (stream devaddr_range_v1);
```

jeffgrunewald (commented):
I’m also wondering if we need to add a verifiable signature to all of the config service’s responses to avoid someone man-in-the-middling the req/res cycle from the server, which would probably mean we’d need some sort of confirmation field in those currently empty success responses on route updates and deletes.
The gateway service’s region params response does this as a carryover from the validators

src/service/iot_config.proto
```diff
@@ -378,19 +386,36 @@ service route {
...
   // Update Route for an Org (auth delegate_keys/owner/admin)
   rpc update(route_update_req_v1) returns (route_v1);
   // Delete Route for an Org (auth delegate_keys/owner/admin)
   rpc delete (route_delete_req_v1) returns (route_v1);
-  // Add / Remove / Update EUIs from a Route (auth delegate_keys/owner/admin)
-  rpc euis(route_euis_req_v1) returns (route_euis_res_v1);
-  // Add / Remove/ Update Devaddr Ranges from a Route (auth
+
+  // Get EUIs for a Route (auth delegate_keys/owner/admin)
+  rpc get_euis(route_get_euis_req_v1) returns (stream eui_pair_v1);
```
michaeldjeffrey, line 396: Is there a way other than comments we can communicate the intention of these endpoints (euis/devaddrs) is to receive all the info for a single route then the server will close the connection?

It might just be me, but seeing a stream returned makes me assume subscription.

src/service/iot_config.proto
```diff
@@ -378,19 +386,36 @@ service route {
   rpc get(route_get_req_v1) returns (route_v1);
   // Create Route for an Org (auth delegate_keys/owner/admin)
   rpc create(route_create_req_v1) returns (route_v1);
+
```
michaeldjeffrey, line 389: Now that we've litted `euis` and `devaddrs` up a level, should we clarify `_route` on these endpoints?

src/service/iot_config.proto
```diff
@@ -22,45 +22,43 @@ import "region.proto";
...
   repeated bytes delegate_keys = 4;
-  // Use req timestamps to validate order/avoid replay
-  uint64 nonce = 5 [ deprecated = true ];
 }
 
 // Device address range, ex: 16#00000001 to 16#0000000A
 message devaddr_range_v1 {
-  uint32 start_addr = 1;
-  uint32 end_addr = 2;
+  string route_id = 1;
```
michaeldjeffrey, line 52: `devaddr_range_v1` is also used to communicate devaddr restrictions for orgs, which is outside the scope of a route.

Do we want to have 2?
`devaddr_range_v1` that applies to routes, and `devaddr_constraint_v1` that is just a start and end? and could evolve to have an `org_id` if we want to go that route?

jeffgrunewald: I remember you asking about a separate `constraint` type and now I think it makes sense here that I understand the domain more and we’ve gone down this route in the refactor

src/service/iot_config.proto
jeffgrunewald: it's probably fine since it's all off the `route` service, but an rpc definition of `iot_config.route/create_route` doesn't read too bad to me, either.

src/service/iot_config.proto
```diff
@@ -263,7 +273,11 @@ message route_stream_req_v1 {
 
 message route_stream_res_v1 {
   action_v1 action = 1;
-  route_v1 route = 2;
+  oneof data {
+    route_v1 route = 2;
+    eui_pair_v1 eui_pair = 3;
+    devaddr_range_v1 devaddr_range = 4;
+  }
```
jeffgrunewald, line 280: wondering if we need to have the config server signing all of these responses to ensure the client has the option to verify the responses haven't been spoofed by some bad actor?

src/service/iot_config.proto
jeffgrunewald: i think proto is too limiting to do much here; we could name long-lived streams `_subscription` or similar to differentiate from a large req/res with a finite limit

src/service/iot_config.proto
```diff
@@ -263,12 +276,17 @@ message route_stream_req_v1 {
 
 message route_stream_res_v1 {
   action_v1 action = 1;
-  route_v1 route = 2;
+  oneof data {
+    route_v1 route = 2;
```
jeffgrunewald, line 280: i think we'll need to still have the `update` action in the `action_v1` enum type so we can handle updates on the routes themselves; i think semantically we can get away with eui pairs and devaddr ranges only ever being added or removed but the routes seem like they need to support `update` as well

michaeldjeffrey: I might be stuck on the way HPR is storing routes, but I'm trying to think of a situation where `update` would behave differently than `add`.

I only see `update` saying "you should already have this", but that's only guaranteed through agreement. Nothing should break if they get an update to something that doesn't exist.
But again, that's only thinking of HPR.

jeffgrunewald: Since the route type has field types that could change (maybe just server fields at this point I think) I could see updating the route to change the server port without making any other changes but if the consumer is going to treat that as a full delete/replace operation I can treat it as an update on the config service side and then broadcast it as an `add` operation as long as the HPR knows how to handle that record being attached to an existing route ID

jeffgrunewald: looking at the structure of a route message, i'm guessing that once a route is created the route_id, net_id, and the oui shouldn't change so we're really only talking about the max_copies field and the json structure of the server record that might update; if it's easier for the HPR to just receive `delete` requests to delete route records from ets and `add` requests to upsert records into ets and overwrite records if that route_id already exists that's easy enough for me

jeffgrunewald (approved):

michaeldjeffrey (approved):

PR #264, opened 2023-01-17
Fixup config/router

madninja:
* Add `GatewayClient` to the `iot_config` namespace for consistent access
* Remove `StateChannelClient` from `router` since it’s no longer used
* Remove the BlockchainVarV1 -> ConfigValue trait from the `local` namespace since it’s no longer used in gateway-rs

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -15,6 +15,7 @@ pub mod services {
 
     pub mod iot_config {
         include!(concat!(env!("OUT_DIR"), "/helium.iot_config.rs"));
+        pub use gateway_client::GatewayClient;
         pub use gateway_server::{Gateway, GatewayServer};
         pub use org_client as config_org_client;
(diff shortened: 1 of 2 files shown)


jeffgrunewald (approved):
this looks good but do we want to wait to remove it until _after_ we've shipped the last gateway-rs embedded within the erlang miner (i.e. after off-chain switch) in case we need to do any unexpected releases without having to work one side or the other off a branch?

madninja:
> this looks good but do we want to wait to remove it until _after_ we've shipped the last gateway-rs embedded within the erlang miner (i.e. after off-chain switch) in case we need to do any unexpected releases without having to work one side or the other off a branch?

gateway-rs has too many large branches pending. I figure if miner needs one we can branch off the last tagged gateway-rs release for minor updates.

PR #265, opened 2023-01-18
Add comment for becon report tx_power

madninja:
The reported tx_power does _not_ include the asserted gain. The verification will need to take the asserted gain into account.


ke6jjj (approved):

madninja:
cc @andymck, for clarity

PR #267, opened 2023-01-20
Packet verifier output 

maplant:


src/lib.rs
```diff
@@ -46,6 +46,10 @@ pub mod services {
         pub use api_server::{Api, ApiServer as Server};
     }
 
+    pub mod packet_verifier {
+        include!(concat!(env!("OUT_DIR"), "/helium.packet_verifier.rs"));
```
madninja, line 50: Follow the convention of the other services and expose `Client`, `Server` and the interface that needs implementing

src/service/packet_verifier.proto
```diff
@@ -0,0 +1,15 @@
...
+
+package helium.packet_verifier;
+
+message valid_packet {
+  uint32 payload_size = 1;
+  bytes gateway = 2;
+  bytes payload_hash = 3;
+}
+
+message invalid_packet {
```
madninja, line 11: no invalid reason to include? A decision was made to have it be invalid so we should declare why to avoid losing the context.

madninja: There will be gaming detection so I expect a number of reasons to crop up over time. Follow the model the iot verifier uses with a proto error enum ?

src/lib.rs
maplant: There is no client and server, it's just a daemon. If that's the case, should I just put this stuff somewhere else?

src/service/packet_verifier.proto
maplant: Done

madninja (approved):

PR #269, opened 2023-01-23
replace valid_poc with poc, include valid & invalid witnesses

andymck:
Linked oracle pr: https://github.com/helium/oracles/pull/306

### DIFF
--- a/src/service/poc_lora.proto
+++ b/src/service/poc_lora.proto
@@ -7,7 +7,6 @@ import "data_rate.proto";
 enum verification_status {
   valid = 0;
   invalid = 1;
-  failed = 2;
 }
 
 enum invalid_participant_side {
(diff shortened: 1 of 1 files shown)


src/service/poc_lora.proto
```diff
@@ -4,30 +4,38 @@ package helium.poc_lora;
 
 import "data_rate.proto";
 
+enum verification_status {
+  valid = 0;
+  invalid = 1;
+  failed = 2;
```
bbalser, line 10: What does `failed` mean in this context?

andymck: the failed variant is used with we are unable to determine if a report is valid or invalid....such as if the node were to be down and we could not retrieve gateway info.  In these cases we retry the associated POC again next run

Failed reports never make it to the external proto structs, as eventually if a report keeps failing it will end up being discarded

bbalser (approved):

madninja:
Note that this will require a bucket _reset_ for the iot bucket

PR #270, opened 2023-01-24
config api for loading region info into config svc

jeffgrunewald:
Adds an admin RPC to the iot_config proto allowing an admin user (signing the request messages with the authorized admin keypair) to load the contents of a `Region.res7.h3idx` file output by the lorawan-h3 `lw-generator`, as well as a collection of `blockchain_region_params_v1` into the appropriate region within the config service.

### DIFF
--- a/src/service/iot_config.proto
+++ b/src/service/iot_config.proto
@@ -370,7 +370,7 @@ message load_region_req_v1 {
   blockchain_region_params_v1 params = 2;
   // Gzip-compressed file content from converting region geojson to a list of h3
   // indexes
-  bytes hex_indexes = 3;
+  optional bytes hex_indexes = 3;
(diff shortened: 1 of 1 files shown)


src/service/iot_config.proto
```diff
@@ -365,6 +365,17 @@ message gateway_region_params_res_v1 {
   bytes signature = 4;
 }
 
+message load_region_req_v1 {
+  region region = 1;
+  blockchain_region_params_v1 params = 2;
```
madninja, line 370: We have to support being able to update the region params separate from the hex_indexes since they update on very different schedules. The proto can support it through the optional/default values for either params or hex_indexes but wanted to make sure you got the requirement for the implementation

madninja (approved):

PR #272, opened 2023-01-26
Allow protoc overriding via env var

JayKickliter:
I've run into instances where I do not want to use the vendored `protoc` for whatever reason. This PR allows you to define`PROTOC` and not use the vendored binaries.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -28,8 +28,7 @@ const MESSAGES: &[&str] = &[
 
 #[cfg(feature = "services")]
 fn main() -> Result<()> {
-    std::env::set_var("PROTOC", protoc_bin_vendored::protoc_bin_path().unwrap());
-
+    maybe_set_protoc();
     tonic_build::configure()
         .build_server(true)
         .build_client(true)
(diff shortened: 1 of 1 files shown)


madninja:
but can we avoid the vendored protoc altogether?

JayKickliter:
preferably we could. about the only way to know would be turn off the vendored protoc and see any there's any fallout

madninja:
> preferably we could. about the only way to know would be turn off the vendored protoc and see any there's any fallout

I think we should

JayKickliter:
@madninja: following up, the Docker image ` rust:latest` still fails with
 
 ```
 #13 100.1   --- stderr
#13 100.1   Error: Custom { kind: Other, error: "protoc failed: service/iot_config.proto: This file contains proto3 optional fields, but --experimental_allow_proto3_optional was not set.\n" }
```

With enough effort, we could find a solution that doesn't involve vendoring protoc, but I'm not sure it's worth it at the moment.

madninja:
> @madninja: following up, the Docker image ` rust:latest` still fails with
> 
> ```
> #13 100.1   --- stderr
> #13 100.1   Error: Custom { kind: Other, error: "protoc failed: service/iot_config.proto: This file contains proto3 optional fields, but --experimental_allow_proto3_optional was not set.\n" }
> ```
> 
> With enough effort, we could find a solution that doesn't involve vendoring protoc, but I'm not sure it's worth it at the moment.

no this is a bug in iot_config.. we shouldn't have optional fields at all in protos

JayKickliter:
Closing in favor of #296

PR #273, opened 2023-01-27
Add new fields to poc_mobile data_transfer_session_req_v1

bbalser:


src/service/poc_mobile.proto
```diff
@@ -158,7 +158,7 @@ message data_transfer_session_req_v1 {
   data_transfer_radio_access_technology radio_access_technology = 4;
   string session_id = 5;
   bytes oui = 6;
-  uint64 imsi = 7;
+  bytes imsi = 7;
```
madninja: document this field as the sha256 of the imsi

madninja (approved):

src/service/poc_mobile.proto
```diff
@@ -156,11 +156,14 @@ message data_transfer_session_req_v1 {
   uint64 upload_bytes = 2;
   uint64 download_bytes = 3;
   data_transfer_radio_access_technology radio_access_technology = 4;
+  string session_id = 5;
+  bytes oui = 6;
```
jeffgrunewald, line 160: this is a different OUI than the integer OUI identifier data transfer users of the IoT network have assigned to them i take it?

taaraora, line 160: Should it be payer_pub_key? I thought this should be the pub key of the router that pays for traffic (which is attached to OUI).

madninja: oui doesn't work in the mobile context since getting one requires acquiring slab of lora addresses right @jeffgrunewald ?

If so this would need to become `bytes payer = 6;`.

taaraora (approved):

PR #275, opened 2023-02-01
make invalid reason standalone, import where required

andymck:


madninja:
This doesn’t seem right? You’re not using it elsewhere and it seems way to generically named for iot poc invalid reasons

andymck:
@madninja draft, see discussion on fishbowl

andymck:
but yeah i dont need it in the core proto, overspill from an earlier version where i updated witness channel to use it.

andymck:
dropping as no longer required

PR #278, opened 2023-02-06
Active / Locked route

macpie:
* netid uint32
* Add locked and active to route

### DIFF
--- a/src/service/iot_config.proto
+++ b/src/service/iot_config.proto
@@ -113,12 +113,15 @@ message route_v1 {
   // UUID
   string id = 1;
   // LoraWan Network ID
-  uint64 net_id = 2;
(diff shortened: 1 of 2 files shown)


michaeldjeffrey (approved):

src/service/iot_config.proto
```diff
@@ -154,7 +157,7 @@ message org_create_roamer_req_v1 {
 
 message org_res_v1 {
   org_v1 org = 1;
-  uint64 net_id = 2;
+  uint32 net_id = 2;
```
jeffgrunewald, line 160: should we add `locked` to the org as well?

jeffgrunewald (approved):

PR #280, opened 2023-02-08
Attach hex_scale and reward_unit to poc witness and receipts

vihu:
This attaches `reward_unit` and `hex_scale` to the poc_witness and poc_receipt proto, so that downstream applications (ETL, explorer) can look it up.

### DIFF
--- a/src/blockchain_poc_core_v1.proto
+++ b/src/blockchain_poc_core_v1.proto
@@ -24,6 +24,13 @@ message blockchain_poc_receipt_v1 {
   // fixed point reward shares added by off-chain poc verifier,
   // propose 2 decimal places
   uint32 reward_shares = 13;
+  // NOTE: reward_unit and hex_scale are ignored when signing the receipt txn
+  // integer representation of a 4-point precision decimal multiplier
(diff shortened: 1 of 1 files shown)


src/blockchain_poc_core_v1.proto
```diff
@@ -49,10 +49,10 @@ message blockchain_poc_witness_v1 {
   // NOTE: reward_unit and hex_scale are ignored when signing the receipt txn
   // integer representation of a 4-point precision decimal multiplier
   // based on the number of witnesses to a poc event
-  uint32 reward_unit = 11;
+  optional uint32 reward_unit = 11;
```
madninja, line 52: you can't do these.. not supported by all but the latest protoc compilers. The "default value" is how optional is done for proto3 primitives

madninja (changes requested):

src/blockchain_poc_core_v1.proto
vihu: Uh, alright then

andymck (approved):

madninja (approved):

PR #283, opened 2023-02-09
Add gateway location RPC

macpie:


src/service/iot_config.proto
```diff
@@ -348,6 +348,14 @@ message load_region_req_v1 {
 
 message load_region_res_v1 {}
 
+message gateway_location_req_v1 {
+  bytes gateway = 1;
+  uint64 timestamp = 2;
```
madninja, line 353: Why timestamp ? This is not a mutating call

src/service/iot_config.proto
```diff
@@ -348,6 +348,14 @@ message load_region_req_v1 {
...
+  bytes signature = 3;
```
madninja, line 354: Who is signing this and how would you know who’s signing?

src/service/iot_config.proto
```diff
@@ -348,6 +348,14 @@ message load_region_req_v1 {
...
+}
+
+message gateway_location_res_v1 { string location = 1; }
+
```
madninja, line 358: Should the config service  sign the response ? The caller should already have the signing pubkey

src/service/iot_config.proto
macpie: This is the same as the other messages only a handful of admin keys have access to this RPC

src/service/iot_config.proto
macpie: We don't in any other message

src/service/iot_config.proto
macpie: avoid replays

jeffgrunewald: this is just an informational "getter" correct? is there a replay attack vector with that?

src/service/iot_config.proto
jeffgrunewald: as i recall we had the config service sign responses to the gateways b/c we couldn't guarantee they were able to travel over a tls connection but in the case of the other network (router, hpr, etc) infrastructure we should be able to enforce a tls connection to the config service so the signature wasn't critical

src/service/iot_config.proto
madninja: perhaps add a comment here so it's not confusing next time someone comes across this?

src/service/iot_config.proto
macpie: well we just don't want anybody to hit the API? otherwise I don't care that much

macpie: Will remove

jeffgrunewald (approved):

PR #286, opened 2023-02-15
Remove protoc dependency

madninja:
This removes the vendored protoc dependency and rely on the host platform to have the right version installed.

The vendored protoc were added to suppport `cross` docker images that did not have the ability to add protoc to the image. This is now possible in cross.

This is an alternate solution to #272


JayKickliter (approved):
This is better than the other PR. I closed it

PR #288, opened 2023-02-23
admin api service

jeffgrunewald:
split out rpcs used to manage the config service itself to group them according to function

### DIFF
--- a/src/service/iot_config.proto
+++ b/src/service/iot_config.proto
@@ -337,7 +337,14 @@ message gateway_region_params_res_v1 {
   bytes signature = 4;
 }
 
-message gateway_load_region_req_v1 {
+message gateway_location_req_v1 {
+  bytes gateway = 1;
+  bytes signature = 2;
+}
+
+message gateway_location_res_v1 { string location = 1; }
+
+message admin_load_region_req_v1 {
   region region = 1;
   blockchain_region_params_v1 params = 2;
   // Gzip-compressed file content from converting region geojson to a list of h3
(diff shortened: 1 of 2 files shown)


src/service/iot_config.proto
```diff
@@ -346,14 +353,25 @@ message gateway_load_region_req_v1 {
...
 
-message gateway_location_req_v1 {
-  bytes gateway = 1;
+message admin_add_key_req_v1 {
+  enum key_type_v1 {
+    operator = 0;
+    packet_router = 1;
+  }
+
+  string pubkey = 1;
```
madninja, line 364: So pub keys are still passed around as bytes and not as strings right?

src/service/iot_config.proto
```diff
@@ -346,14 +353,25 @@ message gateway_load_region_req_v1 {
...
+  key_type_v1 key_type = 2;
+  bytes signature = 3;
```
madninja, line 366: Document this field stating that this request needs to be signed by an existing admin key (vs the pub key being added)

src/service/iot_config.proto
```diff
@@ -346,14 +353,25 @@ message gateway_load_region_req_v1 {
...
+}
+
+message admin_remove_key_req_v1 {
+  string pubkey = 1;
```
madninja, line 370: Same? bytes?

src/service/iot_config.proto
```diff
@@ -346,14 +353,25 @@ message gateway_load_region_req_v1 {
...
   bytes signature = 2;
```
madninja, line 371: Document who's expected to be signing

src/service/iot_config.proto
jeffgrunewald: yeah, i was thinking of this only from the perspective of the operator using this but bytes is consistent across all the other uses of passing around a pubkey and it's easy enough to base64 encode them

jeffgrunewald: fixed [ca3da06](https://github.com/helium/proto/pull/288/commits/ca3da06d51faa7b490e2377cf47c20098e9900c5)

src/service/iot_config.proto
jeffgrunewald: added [ca3da06](https://github.com/helium/proto/pull/288/commits/ca3da06d51faa7b490e2377cf47c20098e9900c5)

src/service/iot_config.proto
madninja: the user passes it to the cli as a public key (b58 string). The cli would then decode it.. you're not base64 encoding keys now are you?!

jeffgrunewald: You’re right, b58, not 64

src/service/iot_config.proto
```diff
@@ -346,14 +353,31 @@ message gateway_load_region_req_v1 {
...
 
-message gateway_load_region_res_v1 {}
+message admin_load_region_res_v1 {}
 
-message gateway_location_req_v1 {
-  bytes gateway = 1;
+message admin_add_key_req_v1 {
+  enum key_type_v1 {
+    // administrator key
+    operator = 0;
```
michaeldjeffrey, line 361: Can we call this `administrator`? 

`operator` has been used in the past to mean many things other than `admin`.

jeffgrunewald: sure thing

jeffgrunewald: [89b4930](https://github.com/helium/proto/pull/288/commits/89b493096d25504180a737e8304ff011a8ce50cb)

michaeldjeffrey (approved):

PR #289, opened 2023-02-23
Add valid_data_transfer_session to packet verifier

maplant:
Invalid sessions are not possible right now

### DIFF
--- a/src/service/packet_verifier.proto
+++ b/src/service/packet_verifier.proto
@@ -24,4 +24,8 @@ message valid_data_transfer_session {
   uint64 upload_bytes = 2;
   uint64 download_bytes = 3;
   bytes payer = 4;
+  // Timestamp in millis of the first file we found a data transfer session in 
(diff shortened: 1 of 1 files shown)


src/service/packet_verifier.proto
```diff
@@ -18,3 +18,9 @@ message invalid_packet {
 enum invalid_packet_reason {
   invalid_packet_reason_insufficient_balance = 0;
 }
+
+message valid_data_transfer_session {
+  bytes pub_key = 1;
+  uint64 upload_bytes = 2;
+  uint64 download_bytes = 3;
```
madninja, line 25: no session_id imsi_hash propagated? It seems impossible to trace back to the reported data transfer req that caused this output?

madninja: If you can't charge for it because the balance fell too low would that not be an "invalid one"?

PR #291, opened 2023-02-28
Add support for price report

vihu:
Summary
----

Adds support for price_report

### DIFF
--- a/build-cpp.sh
+++ b/build-cpp.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 SERVICES=$(echo src/service/{iot_config,downlink,follower,gateway,local,packet_router,poc_lora,poc_mobile,router,state_channel,transaction}.proto)
-MESSAGES=$(echo src/{blockchain_txn,entropy,data_rate,region,mapper}.proto)
+MESSAGES=$(echo src/{blockchain_txn,entropy,data_rate,region,mapper,price_report}.proto)
 SRC_FILES="$SERVICES $MESSAGES"
 
 protoc -I=src --cpp_out=. $SRC_FILES
--- a/build.rs
+++ b/build.rs
@@ -25,6 +25,7 @@ const MESSAGES: &[&str] = &[
     "src/mapper.proto",
     "src/reward_manifest.proto",
     "src/blockchain_region_param_v1.proto",
+    "src/price_report.proto",
 ];
 
 macro_rules! config {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -10,7 +10,7 @@ pub use prost::{DecodeError, EncodeError, Message};
 pub mod services {
     use crate::{
         BlockchainRegionParamsV1, BlockchainTokenTypeV1, BlockchainTxn, DataRate, EntropyReportV1,
-        GatewayStakingMode, MapperAttach, Region, RoutingAddress,
+        GatewayStakingMode, MapperAttach, PriceReportV1, Region, RoutingAddress,
     };
 
     pub mod iot_config {
--- a/src/price_report.proto
+++ b/src/price_report.proto
@@ -1,6 +1,6 @@
 syntax = "proto3";
 
-package helium.price;
+package helium;
 
 import "blockchain_token_type_v1.proto";
 
--- a/src/price_report.proto
+++ b/src/price_report.proto
@@ -1,0 +1,16 @@
+syntax = "proto3";
+
+package helium.price;
+
+import "blockchain_token_type_v1.proto";
+
+message price_report_v1 {
+  /// The price, specified in 1/1000000ths of a cent, i.e. 1$ = 1000000.
+  /// NOTE: Helium blockchain specifies $1 = 10^8, whereas on solana only 10^6
+  /// is supported.
+  uint64 price = 1;
+  /// The timestamp for this report (unix epoch)
+  uint64 timestamp = 2;
+  /// The token type for this report
+  blockchain_token_type_v1 token_type = 3;
+}


src/price_oracle.proto
```diff
@@ -0,0 +1,10 @@
+syntax = "proto3";
+
+package helium;
+
+message price_oracle_resp_v1 {
```
vihu, line 5: @madninja @bbalser lmk if the response for price oracles needs to contains something else to make sure its unique and verifiable?

madninja: if I interpret hashc0de's conversation I believe we'll need a token name/type in there too.. and the request could either request a specific token or just request once and get all of them back.  I guess requesting a specific token type initially (to get like MOBILE -> USD)

src/price_oracle.proto
```diff
@@ -0,0 +1,10 @@
...
+  /// The price
+  uint64 price = 1;
```
madninja, line 7: indicate units

vihu: That would be bones still correct? 1 HNT = 10^8 bones?

src/service/price_oracle.proto
```diff
@@ -0,0 +1,12 @@
+syntax = "proto3";
+
+package helium.price_oracle;
+
+import "price_oracle.proto";
+
+// request price oracle
+message price_oracle_req_v1 {}
```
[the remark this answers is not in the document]
vihu: Do we want to reuse `blockchain_token_type_v1` or have a separate enum `TokenType` within the scope of `price_orace.proto`?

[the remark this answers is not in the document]
vihu: Also, where exactly does the cluster (presumably enum of Devnet, Beta and Mainnet) belong? I can put it within the service/price_oracle.proto presumably?

src/price_oracle.proto
jeffgrunewald: Should it be 10^6 now/going forward to start the transition since we know that’s a limitation on solana?

src/price_oracle.proto
jeffgrunewald: Should they be signed or do we expect trusted communication between the price oracle and the other oracles? If I followed the conversation correctly this service is only an abstraction for the other oracle infrastructure to query?

src/service/price_oracle.proto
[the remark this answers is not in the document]
jeffgrunewald: What’s the reasoning behind the request message being defined separately from the service and response definition? I haven’t seen that done before

src/service/price_oracle.proto
```diff
@@ -0,0 +1,12 @@
...
+
+service price_oracle {
```
jeffgrunewald, line 10: We can obviously add it later but do we foresee the need to have a price streaming feed to be reactive to price changes; such as swings outside of some defined threshold?

src/service/price_oracle.proto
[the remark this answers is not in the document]
jeffgrunewald: Will we need to be able to request “historical” pricing data such as “the price as of 6 hours ago” or do we always want the current data at the request time?

src/price_oracle.proto
vihu: I've updated the comment to reflect 1$ = 10^6

src/service/price_oracle.proto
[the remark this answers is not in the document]
vihu: I was largely following the entropy approach for the req and resp (entropy gives back a report though) but we can move both req/resp in the same service definition if that's more precise/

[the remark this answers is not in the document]
madninja: yes please

src/price_oracle.proto
madninja: I don't currently see a need to sign the response it's independently verifiable on the Solana chain if needed

src/service/price_oracle.proto
madninja: That's what the source of the price feed (pyth)  already does. This service just makes it 1. accessible to oracles without needing to include Solana/Pyth sdks  and 2. more resilient to Solana outages and 3. delivers historical data by putting price reports in a bucket for later analysis/use/graphing

vihu:
@madninja @jeffgrunewald Since we don't really require the messaging/grpc part for this, I was thinking that it may be better to just have the price_report under src folder instead of it being a service? Both `price_oracle_req_v1` and `price_oracle` service can be removed from here.

jeffgrunewald:
> @madninja @jeffgrunewald Since we don't really require the messaging/grpc part for this, I was thinking that it may be better to just have the price_report under src folder instead of it being a service? Both `price_oracle_req_v1` and `price_oracle` service can be removed from here.

So the s3 bucket reports would be the primary interface for other oracles to get pricing data rather than a secodary audit trail with the primary interface being the oracle querying the pricing service directly via grpc? if that's the case, that makes sense to me

vihu:
> > @madninja @jeffgrunewald Since we don't really require the messaging/grpc part for this, I was thinking that it may be better to just have the price_report under src folder instead of it being a service? Both `price_oracle_req_v1` and `price_oracle` service can be removed from here.
> 
> So the s3 bucket reports would be the primary interface for other oracles to get pricing data rather than a secodary audit trail with the primary interface being the oracle querying the pricing service directly via grpc? if that's the case, that makes sense to me

Yes, we are not doing GRPC for price reports as I'm told.

jeffgrunewald (approved):

bbalser (approved):

PR #294, opened 2023-03-01
create org update req

macpie:


src/service/iot_config.proto
```diff
@@ -157,6 +157,16 @@ message org_create_roamer_req_v1 {
   repeated bytes delegate_keys = 6;
 }
 
+message org_update_req_v1 {
+  uint64 oui = 1;
+  oneof update {
+    bytes payer = 2;
+    repeated bytes delegate_keys = 3;
```
jeffgrunewald, line 164: would you simply overwrite the list of delegate keys each time or does this need an `action` attribute as well to be able to add/remove delegate keys?

macpie: Was thinking of overwrite

macpie:
We should also think of adding a `RPC`  to handle adding or removing DevAddr ranges on an org

jeffgrunewald:
handled by https://github.com/helium/proto/pull/326

PR #295, opened 2023-03-01
Add Java options for a correct generation

disk91:
These options helps to generate Java classes for Config service properly

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -28,6 +28,7 @@ const MESSAGES: &[&str] = &[
     "src/reward_manifest.proto",
     "src/blockchain_region_param_v1.proto",
     "src/price_report.proto",
+    "src/hex_boosting.proto",
 ];
 
 macro_rules! config {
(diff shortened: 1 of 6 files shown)


madninja (commented):
could you add a java CI build? That way we'll all know when stuff breaks

madninja (changes requested):
One comment and I'd still like to see a Java CI step to ensure this does not break in future builds

disk91:
Hello, It has been updated and the new files with header have been added

src/blockchain_region_param_v1.proto
```diff
@@ -1,6 +1,9 @@
 syntax = "proto3";
 
 package helium;
+option java_package = "com.helium.grpc";
+option java_outer_classname = "regionparam";
```
jeffgrunewald, line 5: `regionparam` and not `RegionParam`?

jeffgrunewald: or `IotRegionParam` for consistency with the `region.proto`?

devilish66694 (approved):

PR #296, opened 2023-03-02
config service APIs and dc transfer related additions for iot verifier

andymck:
Adds config service APIs required by iot verifier
Extends valid packet to include additional fields required for dc rewards on iot verifier

### DIFF
--- a/src/service/iot_config.proto
+++ b/src/service/iot_config.proto
@@ -371,23 +371,22 @@ message admin_remove_key_req_v1 {
 message admin_key_res_v1 {}
 
 message gateway_info {
-  /// The asserted h3 (string) location of the gateway.
-  // Empty string if the gateway is not found
+  /// The asserted h3 location of the gateway.
   string location = 1;
   bytes address = 2;
-  bytes owner = 3;
-  bool is_full_hotspot = 4;
+  bool is_full_hotspot = 3;
   /// the transmit gain value of the gateway in dbi x 10
   // For example 1 dbi = 10, 15 dbi = 150
-  int32 gain = 5;
-  region region = 6;
+  int32 gain = 4;
+  region region = 5;
+  int32 elevation = 6;
 }
 
 /// Look up the details of a given hotspot public key
 message gateway_info_req_v1 {
   /// The pubkey_bin address of the gateway to look up
   bytes address = 1;
-  // sig from a key known to the config service
+  /// sig from a key known to the config service
   bytes signature = 2;
 }
 
@@ -395,32 +394,37 @@ message gateway_info_res_v1 {
   /// Timestamp of response
   uint64 timestamp = 1;
   gateway_info info = 2;
-  // sig from the config service
+  /// sig from the config service
   bytes signature = 3;
 }
 
 /// Request a stream of all active gateways
 message gateway_info_stream_req_v1 {
   uint32 batch_size = 1;
-  // sig from a key known to the config service
+  /// sig from a key known to the config service
   bytes signature = 2;
 }
 
 /// Active gateway info streaming response containing a batch of gateways
 message gateway_info_stream_res_v1 {
-  repeated gateway_info_res_v1 gateways = 1;
+  /// Timestamp of response
+  uint64 timestamp = 1;
+  /// batch of gateways
+  repeated gateway_info gateways = 2;
+  /// sig from the config service
+  bytes signature = 3;
 }
 
 message region_params_req_v1 {
   region region = 1;
-  // sig from a key known to the config service
+  /// sig from a key known to the config service
   bytes signature = 2;
 }
 
 message region_params_res_v1 {
   region region = 1;
   blockchain_region_params_v1 params = 2;
-  // sig from the config service
+  /// sig from the config service
   bytes signature = 3;
 }
 
--- a/src/service/iot_config.proto
+++ b/src/service/iot_config.proto
@@ -381,13 +381,13 @@ message gateway_info {
   // For example 1 dbi = 10, 15 dbi = 150
   int32 gain = 5;
   region region = 6;
-  blockchain_region_params_v1 region_params = 7;
 }
 
 /// Look up the details of a given hotspot public key
 message gateway_info_req_v1 {
   /// The pubkey_bin address of the gateway to look up
   bytes address = 1;
+  // sig from a key known to the config service
   bytes signature = 2;
 }
 
@@ -395,18 +395,33 @@ message gateway_info_res_v1 {
   /// Timestamp of response
   uint64 timestamp = 1;
   gateway_info info = 2;
+  // sig from the config service
   bytes signature = 3;
 }
 
 /// Request a stream of all active gateways
 message gateway_info_stream_req_v1 {
   uint32 batch_size = 1;
+  // sig from a key known to the config service
   bytes signature = 2;
 }
 
 /// Active gateway info streaming response containing a batch of gateways
 message gateway_info_stream_res_v1 {
   repeated gateway_info_res_v1 gateways = 1;
+}
+
+message region_params_req_v1 {
+  region region = 1;
+  // sig from a key known to the config service
+  bytes signature = 2;
+}
+
+message region_params_res_v1 {
+  region region = 1;
+  blockchain_region_params_v1 params = 2;
+  // sig from the config service
+  bytes signature = 3;
 }
 
 // ------------------------------------------------------------------
@@ -497,4 +512,6 @@ service admin {
   // Load params and cell indexes for a region into the config service (auth
   // admin only)
   rpc load_region(admin_load_region_req_v1) returns (admin_load_region_res_v1);
+  // Return the region params for the specified region
+  rpc region_params(region_params_req_v1) returns (region_params_res_v1);
 }
--- a/src/service/iot_config.proto
+++ b/src/service/iot_config.proto
@@ -370,6 +370,45 @@ message admin_remove_key_req_v1 {
 
 message admin_key_res_v1 {}
 
+message gateway_info {
+  /// The asserted h3 (string) location of the gateway.
+  // Empty string if the gateway is not found
+  string location = 1;
+  bytes address = 2;
+  bytes owner = 3;
+  bool is_full_hotspot = 4;
+  /// the transmit gain value of the gateway in dbi x 10
+  // For example 1 dbi = 10, 15 dbi = 150
+  int32 gain = 5;
+  region region = 6;
+  blockchain_region_params_v1 region_params = 7;
+}
+
+/// Look up the details of a given hotspot public key
+message gateway_info_req_v1 {
+  /// The pubkey_bin address of the gateway to look up
+  bytes address = 1;
+  bytes signature = 2;
+}
+
+message gateway_info_res_v1 {
+  /// Timestamp of response
+  uint64 timestamp = 1;
+  gateway_info info = 2;
+  bytes signature = 3;
+}
+
+/// Request a stream of all active gateways
+message gateway_info_stream_req_v1 {
+  uint32 batch_size = 1;
+  bytes signature = 2;
+}
+
+/// Active gateway info streaming response containing a batch of gateways
+message gateway_info_stream_res_v1 {
+  repeated gateway_info_res_v1 gateways = 1;
+}
+
 // ------------------------------------------------------------------
 // Service Definitions
 // ------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


src/service/iot_config.proto
```diff
@@ -365,6 +365,53 @@ message admin_remove_key_req_v1 {
...
+
+enum gateway_error_reason {
+  reason_none = 0;
+  gateway_not_found = 1;
+}
+
+message gateway_error_res {
+  bytes address = 1;
+  gateway_error_reason error = 2;
+}
```
jeffgrunewald, line 390: we should be able to omit these; tonic can return a more standard http/grpc-level error response with the `Status` type, `not_found` being one of them. i would just make the response message type be the gateway_info struct, the timestamp, and we'll need a signature i would imagine since these will no longer be host-local the way the blockchain-node was running

src/service/iot_config.proto
```diff
@@ -365,6 +365,53 @@ message admin_remove_key_req_v1 {
...
+
+/// Look up the details of a given hotspot public key
+message gateway_info_req_v1 {
+  /// The pubkey_bin address of the gateway to look up
+  bytes address = 1;
```
jeffgrunewald, line 395: this may need a request signature depending on how we want to setup the network around the config service and between it and the other oracles; we may also need a timestamp on the request for the purposes of rate-limiting potential non-oracle users. thoughts @madninja ?

src/service/iot_config.proto
```diff
@@ -365,6 +365,53 @@ message admin_remove_key_req_v1 {
...
+}
+
+message gateway_info_res_v1 {
+  /// Timestamp of response
+  uint64 timestamp = 1;
+  oneof result {
+    gateway_info info = 2;
+    gateway_error_res error = 3;
```
jeffgrunewald, line 403: given the above suggestion about using tonic/grpc status responses to convey errors i would pull the `gateway_info` back up to the top level here outside of any `oneof` and again depending on our network topography this will need a signature to verify by the oracle as well

src/service/iot_config.proto
madninja: 1. Timestamp isn't going to help with rate limiting
2. Signing the request and response sounds like a good idea.. We should talk through whether 
    - This info service is open to the public, in which case only the response needs to be signed by the config service so the caller can confirm it came from the config service
    - This info service is private to a set of approved keys, in which case the request is signed by the caller and verified against approved keys. The  response is signed by the config service

I believe we should limit access since listing hotspots is a wildly popular script

src/service/iot_config.proto
```diff
@@ -438,6 +485,11 @@ service gateway {
       returns (gateway_region_params_res_v1);
   // Get H3 Location for a gateway (auth admin only)
   rpc location(gateway_location_req_v1) returns (gateway_location_res_v1);
+  // Get info for the specified gateway
+  rpc info(gateway_info_req_v1) returns (gateway_info_res_v1);
+  // Get a stream of gateway info
+  rpc info_stream(gateway_info_stream_req_v1)
+      returns (stream gateway_info_stream_res_v1);
```
jeffgrunewald, line 492: do we want or is there any advantage in having the info stream return a different/reduced message type than the batches of the full `gateway_info` when it's only being consumed to build the transmit scaling map and that only needs the `location` field? especially since we just changed the blockchain-node implementation to return an empty region_params field for this request/response type?
i _believe_ we could change this to just return a stream of batches of `gateway_location_res_v1` if all we expect to need is to maintain compatibility with how the verifier is using the current node but we could also have a separate `gateway_metadata_res_v1` that just includes the location and other fields we might need for _all_ gateways

src/service/iot_config.proto
andymck: grpc protocol level error msgs are limited in the context they provide.  Perhaps they are sufficient for the purpose of the config service but if responses in any way need to be attestable  then application level error msgs may be required.  Something to keep in mind

src/service/iot_config.proto
andymck: so i am hoping to be able to use the stream response to prewarm the gateway cache in the verifier.  That will save considerable time after a reboot.  Currently we await ~5 mins to stream the gateways to the density scaler and then ~20mins to complete the first batch of beacons and witnesses.  Prewarming the cache will reduce this time significantly

src/service/iot_config.proto
andymck: addressed here: e8f3a601a1ff0713430ee9e1b70406f3efeb39ff

src/service/iot_config.proto
andymck: addressed here: e8f3a601a1ff0713430ee9e1b70406f3efeb39ff

src/service/iot_config.proto
andymck: addressed here: e8f3a601a1ff0713430ee9e1b70406f3efeb39ff

src/service/iot_config.proto
jeffgrunewald: Does that include the region params or can those still be omitted for that cache? If you need those region params then we’ll likely still have to solve for some significant memory pressure on the config service without tying the cached region params on _that_ side into the info stream

andymck: Given region params dont change all that often and there is a small set of these, we can probably do without and pull those from a local verifier cache based on the gateway.region field.

andymck: so we can likely drop the region params from the gateway info struct

andymck: I suspect an RPC to the config service to return the region params for a specified region will be the best fit here.  Will confirm after working through some things on the verifier side

andymck: I see there is an existing `region_params` RPC.  That works for my needs

jeffgrunewald: that one may not be the best bet if you want to load the params once and cache them since that rpc takes the pubkey of a hotspot in the request (and the hotspot’s “default” region) and looks up the pubkey on-chain to return the asserted region’s params

andymck: Added this: 4212e57086fb29c05e71e9c0e48bc2c034f9d877 to return region params for the specified region.  Its very close to the GatewayRegionParamsReqV1, is that RPC actually in use atm ?

madninja: I'm not sure I like the way this is coming together right now.

andymck: in what sense? specifics would be helpful

jeffgrunewald: in this case i would expect the signing key for the verifier, or any other oracle, to have its pubkey registered to the config service (perhaps under a new key type of "oracle" as opposed to the "administrator" or "packet router" types) using the admin service's `add_key` rpc. for any of these methods we want to truly authorize we have to save the key in the config service somewhere and retrieve it to check the request is authorized anyway so sending the pubkey in the request doesn't seem to add any value to me.

jeffgrunewald: having requests coming from the gateways themselves include their pubkeys for signature validation make sense on the other hand since there's enough of those to make persisting them infeasible and we're not using those as an authorization gate to restrict access to the data

src/service/iot_config.proto
```diff
@@ -365,6 +365,60 @@ message admin_remove_key_req_v1 {
...
+message gateway_info_stream_req_v1 {
+  uint32 batch_size = 1;
+  // sig from a key known to the config service
+  bytes signature = 2;
+}
+
+/// Active gateway info streaming response containing a batch of gateways
+message gateway_info_stream_res_v1 {
+  repeated gateway_info_res_v1 gateways = 1;
+}
```
jeffgrunewald, line 407:
```suggestion
message gateway_info_stream_res_v1 {
  repeated gateway_info gateways = 1;
  uint64 timestamp = 2;
  bytes signature = 3;
}
```
i know this seems redundant with the existing `gateway_info_res_v1` but since the info stream returns a stream of this message type as responses does it make sense to lift the metadata of the request up to the response-level message? when we validate the signature on these streaming responses we already need to grab the first message from the stream and verify the signature and with this structure we'd need to grab the first element of the batch from the first message in the stream to verify and it will probably be more performant to grab one timestamp and get one signature from the config service per batch as opposed to timestamping and signing for every one of the 500-700K gateways this api returns

src/service/iot_config.proto
```diff
@@ -365,6 +365,60 @@ message admin_remove_key_req_v1 {
 
 message admin_key_res_v1 {}
 
+message gateway_info {
+  /// The asserted h3 (string) location of the gateway.
+  // Empty string if the gateway is not found
+  string location = 1;
+  bytes address = 2;
+  bytes owner = 3;
+  bool is_full_hotspot = 4;
```
jeffgrunewald, line 374: looking through the helium-progam-library, including the `helius-service` package i'm reminded that post-Solana migration hotspots won't have `owner` any longer so this field should probably be removed from the metadata in the `gateway_info` message type.
the helius service (and the general `IotHotspotV0` metadata about iot gateways in general) doesn't indicate it will have any concept of `full` vs. what we currently call a `data-only` hotspot so i don't know where this will come from. it _also_ doesn't have a concept of tracking the hotspot's `region` but the config service can compute that from the on-chain `location` we get back from helius so that shouldn't be a problem.

helius _will_ also have the hotspot's asserted `elevation` which we don't currently use but wonder if we want to include here as well in a "reserved for future use" capacity since it's just an `Option<i32>` it would be cheap to include in case it were ever needed for any PoC validations

madninja: * yes on removing owner. No such thing in off-chain oracles
* we need to fix the full/data-only hole with the Solana folks.. That's a gap
* yes region is _always_ derived from the asserted location, and never a stored metadata attribute
* yes the information should includes all asserted hotspot metadata:  location, height, gain

andymck: i actually had this all done in my previous helius PR which i subsequently closed...added elevation, removed owner here: feda3bb55f149a587a7a7f343d506fb1df6dd7d6

andymck: We should also reconsider `location`, currently it is a string which is a carry over from node...unless there is good to keep as string ( which i dont know off ) this should revert to u64 as per solana program libs

madninja: Not really a fan of that right now.. It'd 1. be inconsistent with the other protos and 2. the solana program libs represent it as a string of a u64 which is very bleh compared to the h3 to_string version which is how humans read it

andymck: are you saying you want to keep it as a string or change to u64 ? solana program libs have it defined as a plain old u64, not a string of a u64.  Perhaps its translates it to a string of a u64 at some point, but def defined as a u64

madninja: oh you mean Solana rust program libs? or typescript? Because a u64 translates to a string in json form. 

Either way, dunno specifics and not against making this change, but if we do this it's going to be inconsistent with other protos that include location? If so I'd prefer consistency

andymck: yeah so solana rust program libs declare location as a u64. It was same on the current l1 ledger where it is an integer.

It was changed to a string as part of the follower RPCs

The verifier internally represents location as a u64, converts the proto value from string to u64.  

But i see your point, changing it on the proto side now does mean there is an inconsistency in files on S3 etc so prob best to keep as is. Verifier will continue to represent internally as u64 anyway

jeffgrunewald: fwiw the string makes it easy when parsing _into_ a u64 to attempt to parse the hexadecimal h3.to_string() call and convert the result to an option, where the empty string neatly converts to `None`. if we made the proto version a u64 the default would be 0 which we'd probably need to explicitly check for and perhaps then separately try to parses the raw integer into an H3 cell to check it as a valid h3 index vs. just some garbage non-zero integer

jeffgrunewald (approved):

src/service/poc_lora.proto
```diff
@@ -180,6 +180,8 @@ message gateway_reward_share {
   uint64 start_period = 4;
   /// Unix timestamp in seconds of the end of the reward period
   uint64 end_period = 5;
+  /// Amount credited to the hotspot for data transfer
+  uint64 dc_amount = 6;
```
jeffgrunewald, line 184: Nit: `dt_amount` or `transfer_amount` since the hotspots aren’t receiving or transferring actual data credits; this could be confusing for an external integrator/auditor

bbalser (approved):

src/service/iot_config.proto
```diff
@@ -370,6 +370,64 @@ message admin_remove_key_req_v1 {
...
+/// Look up the details of a given hotspot public key
+message gateway_info_req_v1 {
+  /// The pubkey_bin address of the gateway to look up
+  bytes address = 1;
+  /// sig from a key known to the config service
+  bytes signature = 2;
+}
+
+message gateway_info_res_v1 {
+  /// Timestamp of response
```
madninja, line 394: units

src/service/iot_config.proto
```diff
@@ -370,6 +370,64 @@ message admin_remove_key_req_v1 {
...
+message gateway_info_stream_req_v1 {
+  uint32 batch_size = 1;
+  /// sig from a key known to the config service
+  bytes signature = 2;
+}
+
+/// Active gateway info streaming response containing a batch of gateways
+message gateway_info_stream_res_v1 {
+  /// Timestamp of response
+  uint64 timestamp = 1;
```
madninja, line 411: timestamp units

PR #297, opened 2023-03-06
first revision of mobile config service proto

jeffgrunewald:
define the initial implementation of a `mobile-config` service, the mobile subdao equivalent to the iot-config service that supplies various network configuration and metadata access to the iot subdao data oracles.

the first draft of the mobile config service provides the following three service definitions and their associated message types:
- hotspot service: retrieves individual metadata for a given hotspot, requested by the hotspot's NFT identifying pubkey binary, a successful record retrieval indicates the presence of the request hotspot on-chain. initial hotspot metadata includes the pubkey binary and the hotspot's asserted h3 location index. also includes a "list all" streaming rpc for getting all on-chain hotspots
- router service: confirm the authority of a given router to burn data credits before doing so by verifying the presence of the router's pubkey binary registered with the config service. also includes a "list all" unary rpc for retrieving the public keys of all routers registered to the config service with burn authority to transfer data packets
- admin service: manage registration of authorizing keys to the config service, as well as future management apis to be defined

### DIFF
--- a/src/service/mobile_config.proto
+++ b/src/service/mobile_config.proto
@@ -9,9 +9,10 @@ package helium.mobile_config;
 // == Field Notes ==
 //
 // - Every message including a `signature` field will need to be signed over the
-//   complete message, with the `signature` field set to an empty value.
-//
-// - Every `timestamp` is in milliseconds since unix epoch
+//   complete message, with the `signature` field set to an empty value. Requests
+//   are expected to be signed by the client to validate authorization to make
+//   the request and responses are signed by the mobile_config server to ensure
+//   validity of the data returned
 //
 // - Keypair fields are binary encoded public keys, Rust encoding example here:
 //    https://github.com/helium/helium-crypto-rs/blob/main/src/public_key.rs#L347-L354
--- a/src/service/mobile_config.proto
+++ b/src/service/mobile_config.proto
@@ -33,11 +33,13 @@ message hotspot_metadata_req_v1 {
 
 message hotspot_metadata_res_v1 {
   hotspot_metadata metadata = 1;
+  // unix epoch timestamp in seconds
   uint64 timestamp = 2;
   bytes signature = 3;
 }
 
 message hotspot_metadata_stream_req_v1 {
+  // max number of hotspot metadata in each message of the response stream
   uint32 batch_size = 1;
   bytes signature = 2;
 }
@@ -45,6 +47,7 @@ message hotspot_metadata_stream_req_v1 {
 message hotspot_metadata_stream_res_v1 {
   // a list of hotspot metadata numbering up to the request batch size
   repeated hotspot_metadata hotspots = 1;
+  // unix epoch timestamp in seconds
   uint64 timestamp = 2;
   bytes signature = 3;
 }
@@ -59,6 +62,7 @@ message router_get_req_v1 {
 message router_get_res_v1 {
   // Pubkey binary of the requested router, confirming registration
   bytes pubkey = 1;
+  // unix epoch timestamp in seconds
   uint64 timestamp = 2;
   // Signature over the response by the config service
   bytes signature = 3;
@@ -72,6 +76,7 @@ message router_list_req_v1 {
 message router_list_res_v1 {
   // List of public key binaries of all registered routers
   repeated bytes routers = 1;
+  // unix epoch timestamp in seconds
   uint64 timestamp = 2;
   // Signature over the response by the config service
   bytes signature = 3;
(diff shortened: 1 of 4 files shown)


bbalser (approved):

src/service/mobile_config.proto
```diff
@@ -0,0 +1,129 @@
...
+  repeated bytes routers = 1;
+  uint64 timestamp = 2;
+  // Signature over the response by the config service
+  bytes signature = 3;
+}
+
+message admin_add_key_req_v1 {
+  enum key_type_v1 {
+    // administrative operator key
+    administrator = 0;
```
andymck, line 82: hmm doesnt field numbering start at 1 ?

jeffgrunewald: this is an enum embedded within the message which we've tended to do with really specific enums that don't need to be reused across multiple message types

andymck: ah yes its a enum, ignore

jeffgrunewald:
Some things I noted reading through the Mobile subDAO HIPs that reference PoC including the "Approved" HIP 74, are we interested in including any of the following data in the model:
- CPI registration info?
- Indoor vs. Outdoor? (this affects what coverage is _expected_ from a given hotspot in the mobile poc algorithm)
- Orientation/direction? (this is indicated to factor into the HIP 74 algo proposal)
- Transmit power? (similar to gain; used by HIP 74 to confirm expected vs. observed coverage in neighboring hexes to the asserted location)
- Spectrum access request/grant time?

src/service/mobile_config.proto
```diff
@@ -0,0 +1,129 @@
...
+package helium.mobile_config;
+
+// ------------------------------------------------------------------
+// Message Definitions
+// ------------------------------------------------------------------
+
+// == Field Notes ==
+//
+// - Every message including a `signature` field will need to be signed over the
+//   complete message, with the `signature` field set to an empty value.
```
madninja, line 12: Be clear about req vs res signing and who is doing the signing

src/service/mobile_config.proto
```diff
@@ -0,0 +1,129 @@
...
+//
+// - Every `timestamp` is in milliseconds since unix epoch
```
madninja, line 14: Why millis? for metadata that's overkill

madninja:
> Some things I noted reading through the Mobile subDAO HIPs that reference PoC including the "Approved" HIP 74, are we interested in including any of the following data in the model:
> 
> * CPI registration info?
> * Indoor vs. Outdoor? (this affects what coverage is _expected_ from a given hotspot in the mobile poc algorithm)
> * Orientation/direction? (this is indicated to factor into the HIP 74 algo proposal)
> * Transmit power? (similar to gain; used by HIP 74 to confirm expected vs. observed coverage in neighboring hexes to the asserted location)
> * Spectrum access request/grant time?

We'll need to go through this with zer0tweets and co and figure out what's needed when. The initial launch does not need any of it.. and then we'd need to figure out if this comes through the config service or through ingested data

src/service/mobile_config.proto
jeffgrunewald: removed; it looks like we typically call `to_timestamp_millis()` about twice as often as we call `to_timestamp()` throughout the oracles repo but that's still a fair number of non-milli uses; we'll leave this up to the calling context to determine

madninja: huh? Not clear what removing means here, but the timestamp unit needs to be relevant to the context indeed. The key is that unlike for PoC config changes aren't millisecond level time critical so keeping them at their right unit is what we're shooting for. 

And documenting them when/where used. I actually prefer documenting them in the message itself to avoid people not reading the top of the file

jeffgrunewald: fixed [05ef4df](https://github.com/helium/proto/pull/297/commits/05ef4df63be35fb6a973182cf8641d2d0c8f8765)

src/service/mobile_config.proto
jeffgrunewald: fixed [05ef4df](https://github.com/helium/proto/pull/297/commits/05ef4df63be35fb6a973182cf8641d2d0c8f8765)

src/service/mobile_config.proto
madninja: `timestamp` is fine with a comment right above it as to what unit it is.. Convention has been to not put types in the field names as much as possible

jeffgrunewald: i misunderstood the ask to document the unit "in the message"; commented [fc95f16](https://github.com/helium/proto/pull/297/commits/fc95f16cc1cf65a605c4362eb5b3395fe1c79216)

PR #303, opened 2023-03-16
Add router_status rpc call

madninja:
Adds a `router` status local GRPC call to get the current connection status of the packet router


JayKickliter (approved):
I'm approving since this is so simple, but do you want to test gwrs side of this before merging? It's up to you

madninja:
> I'm approving since this is so simple, but do you want to test gwrs side of this before merging? It's up to you

Already tested on a local branch.. PR coming

PR #312, opened 2023-04-05
update proto for non gateway related rewards

andymck:
Refactors the GatewayRewardShare into IotRewardShare and allows for various reward types to be supported, each with their own individual data requirements

### DIFF
--- a/src/service/poc_lora.proto
+++ b/src/service/poc_lora.proto
@@ -167,21 +167,32 @@ message lora_poc_v1 {
   repeated lora_verified_witness_report_v1 unselected_witnesses = 4;
 }
 
-// iot per-gateway reward share for a given reward period
-// for PoC coverage events provided between the start and end timestamps
-message gateway_reward_share {
+message gateway_reward {
   /// Public key of the hotspot
   bytes hotspot_key = 1;
   /// Amount credited to the hotspot for beaconing
   uint64 beacon_amount = 2;
   /// Amount credited to the hotspot for witnessing
   uint64 witness_amount = 3;
-  /// Unix timestamp in seconds of the start of the reward period
-  uint64 start_period = 4;
-  /// Unix timestamp in seconds of the end of the reward period
-  uint64 end_period = 5;
   /// Amount credited to the hotspot for data transfer
-  uint64 dc_transfer_amount = 6;
+  uint64 dc_transfer_amount = 4;
+}
+
+message operational_reward {
+  /// Amount credited to the operational fund wallet
+  uint64 amount = 1;
+}
+
+message iot_reward_share {
+  /// Unix timestamp in seconds of the start of the reward period
+  uint64 start_period = 1;
+  /// Unix timestamp in seconds of the end of the reward period
+  uint64 end_period = 2;
+  /// the reward allocations for this share
+  oneof reward {
+    gateway_reward gateway_reward = 3;
+    operational_reward operational_reward = 4;
+  }
 }
 
 service poc_lora {
--- a/src/service/poc_lora.proto
+++ b/src/service/poc_lora.proto
@@ -170,16 +170,16 @@ message lora_poc_v1 {
 message gateway_reward {
   /// Public key of the hotspot
   bytes hotspot_key = 1;
-  /// Amount credited to the hotspot for beaconing
+  /// Amount in iot bones credited to the hotspot for beaconing
   uint64 beacon_amount = 2;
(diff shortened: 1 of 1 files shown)


src/service/poc_lora.proto
```diff
@@ -167,8 +167,14 @@ message lora_poc_v1 {
   repeated lora_verified_witness_report_v1 unselected_witnesses = 4;
 }
 
+enum reward_type {
+  gateway = 0;
+  operational = 1;
+  oracles = 2;
```
madninja, line 173: Pick plural or singular.. either `gateways` and `oracles`  or `gateway` and `oracle`. Since this talks about a single reward I suggest singular

src/service/poc_lora.proto
```diff
@@ -180,8 +186,16 @@ message gateway_reward_share {
   uint64 start_period = 4;
   /// Unix timestamp in seconds of the end of the reward period
   uint64 end_period = 5;
+  /// type of reward
+  reward_type reward_type = 6;
```
madninja, line 190: You can't just insert a field and renumber. That'll break all the existing files. You will need to add a number at the end.

src/service/poc_lora.proto
```diff
@@ -180,8 +186,16 @@ message gateway_reward_share {
...
   /// Amount credited to the hotspot for data transfer
-  uint64 dc_transfer_amount = 6;
+  uint64 dc_transfer_amount = 7;
```
madninja, line 192: I'm not sure I like this approach since you now have both a type and a separate amount for each. Why not rename dc_transfer_amount to amount and just have reward_type indicate who that reward goes to and what the amount "means/unit"?

src/service/poc_lora.proto
andymck: hmm  nothing is renumbered that was already in use

src/service/poc_lora.proto
andymck: dc_transfer amount is not related to these current changes for non gateway rewards.  The proto already defines the following:
beacon_amount
witness_amount
dc_transfer_amount

So we then need a new field or fields to cover the non gateway rewards ( operational fund, stakers and oracles ).  The choice here was to have either a single field to cover all 3 or a distinct field for each.   I started with a singular field to cover the 3 new values,  however as we already have a convention whereby we have specific fields per reward type ie  beacon rewards, witness rewards and dc transfer rewards i felt reusing the one field for a subset of rewards was a poor convention.

src/service/poc_lora.proto
madninja: Screenshot 2023-04-05 at 7 44 31 AM

no? Am I not seeing this right?

src/service/poc_lora.proto
andymck: The enum provides a simple means for the code ( indexer ) to identify which fields to use without relying on checking for non zero values for each `_amount` field

madninja: bah.. not a great design choice at the time to have separate amounts there in hindsight. 

If we have separate amounts then we wouldn't need the reward_type at all do we?

madninja: > The enum provides a simple means for the code ( indexer ) to identify which fields to use without relying on checking for non zero values for each `_amount` field

Right but we have to do that for the other amounts already?

src/service/poc_lora.proto
andymck: fixed, all now singular

src/service/poc_lora.proto
andymck: dc_transfer_amount is not yet in use, nothing has been deployed with that

src/service/poc_lora.proto
```diff
@@ -167,21 +167,32 @@ message lora_poc_v1 {
...
   /// Amount credited to the hotspot for beaconing
   uint64 beacon_amount = 2;
   /// Amount credited to the hotspot for witnessing
   uint64 witness_amount = 3;
+  /// Amount credited to the hotspot for data transfer
+  uint64 dc_transfer_amount = 4;
+}
+
+message operational_reward {
+  /// Amount credited to the operational fund wallet
```
madninja, line 182: maybe declare unit here for the amount?

andymck: addressed here: 870adbada95b32010d42db966a1dbc55b1b73605

madninja (approved):
Assuming it matches the mobile one this looks good

bbalser (approved):

PR #313, opened 2023-04-05
Add new messages for mobile rewards

maplant:


src/service/poc_mobile.proto
```diff
@@ -115,6 +115,8 @@ message speedtest {
   uint64 timestamp = 4;
 }
 
+/// Deprecated reward message for mobile. See mobile_reward_share and
```
madninja, line 118: Actually declare as deprecated.. see https://stackoverflow.com/questions/52781727/how-can-i-deprecate-whole-message-in-protocol-buffers

maplant: good catch! Thanks

madninja:
You'll also want to remove the unused ones.. follow @bbalser lead here

src/service/poc_mobile.proto
```diff
@@ -130,6 +131,30 @@ message radio_reward_share {
   uint64 end_epoch = 6;
 }
 
+message radio_reward {
+  /// Public key of the owner of the radio
+  bytes owner_key = 1;
```
bbalser, line 136: Do we need `owner_key` for the new message?  Once we hit solana, the owner_key is no longer used.

maplant: removed

bbalser (approved):

PR #314, opened 2023-04-06
Add an "unknown" region

madninja:
Adds an UNKNOWN region to allow for an unknown region request for a gateway


jeffgrunewald (commented):
does this also need a `FromStr` implementation in the block above the `Display` impl or do we not want to be able to create this variant from a string like the others?

madninja:
> does this also need a `FromStr` implementation in the block above the `Display` impl or do we not want to be able to create this variant from a string like the others?

I wasn't going to create it since you really shouldn't ever see it in string form

PR #316, opened 2023-04-07
Add gateway to packet_router_packet_down_v1

macpie:


src/service/packet_router.proto
```diff
@@ -71,6 +71,7 @@ message packet_router_packet_down_v1 {
   bytes payload = 1;
   window_v1 rx1 = 2;
   window_v1 rx2 = 3;
+  bytes gateway = 4;
```
madninja, line 74: Why add this overhead to every packet headed to the gateway?

macpie: No, this is an experiment to handle data from Router to HPR with less streams.

madninja: Gotcha but it’s also the proto message that the gateway receives right? If so you may want to put the downlink packet in an envelope with the gateway in the outer envelope

macpie: We could leave it empty when it goes to gateway

madninja: Seems like that breaks some system boundaries in favor of some optimization. Let’s chat this through before nailing it down

devilish66694 (approved):

PR #319, opened 2023-04-18
Put SKF under Routes

macpie:


src/service/iot_config.proto
```diff
@@ -653,22 +659,18 @@ service route {
...
+  rpc get_skfs(route_skf_get_req_v1) returns (stream skf_v1);
   // Update Filters for an Org (auth delegate_keys/owner/admin)
-  rpc update(stream session_key_filter_update_req_v1)
-      returns (session_key_filter_update_res_v1);
+  rpc update_skfs(stream route_skf_update_req_v1)
+      returns (route_skf_update_res_v1);
   // Stream Filter update (auth admin only)
-  rpc stream(session_key_filter_stream_req_v1)
-      returns (stream session_key_filter_stream_res_v1);
+  rpc stream_skfs(route_skf_stream_req_v1)
```
jeffgrunewald, line 672: we're moving the session key filters rpcs out of their own service and underneath the route service wouldn't this stream of updates just get combined with the route stream which currently outputs route, eui_pair, and devaddr_range updates and toggles based on the `oneof data` type in the `route_stream_res_v1` message?

src/service/iot_config.proto
```diff
@@ -368,24 +368,24 @@ message route_stream_res_v1 {
...
+message skf_v1 {
+  string route_id = 1;
   uint32 devaddr = 2;
   // the hex-encoded string of the binary session key
   string session_key = 3;
 }
 
-message session_key_filter_list_req_v1 {
-  uint64 oui = 1;
+message route_skf_list_req_v1 {
```
jeffgrunewald, line 378: lists all skfs attached to a single route, correct?

src/service/iot_config.proto
```diff
@@ -368,24 +368,24 @@ message route_stream_res_v1 {
...
   // in milliseconds since unix epoch
   uint64 timestamp = 2;
   bytes signature = 3;
   // pubkey binary of the signing keypair
   bytes signer = 4;
 }
 
-message session_key_filter_get_req_v1 {
-  uint64 oui = 1;
+message route_skf_get_req_v1 {
```
jeffgrunewald, line 387: "list" all the skfs attached to a single devaddr attached to a single route, correct?

src/service/iot_config.proto
macpie: Yes

src/service/iot_config.proto
macpie: yes

src/service/iot_config.proto
macpie: Done

jeffgrunewald (approved):

PR #324, opened 2023-05-02
Add new fields to packet_router_packet_report_v1

macpie:


src/service/packet_router.proto
```diff
@@ -21,6 +26,8 @@ message packet_router_packet_report_v1 {
   bytes payload_hash = 10;
   uint32 payload_size = 11;
   bool free = 12;
+  packet_type type = 13;
+  uint64 hpr_timestamp_ms = 14;
```
madninja, line 30: rename to `receive_timestamp`. In general we don't put the type in the name but put it as a comment above the field.. I.e. 

```
// received timestamp  in ms since unix epoch
uint64 received_timestamp = 14;
```

src/service/packet_router.proto
```diff
@@ -6,6 +6,11 @@ import "region.proto";
 import "data_rate.proto";
 
 message packet_router_packet_report_v1 {
+  enum packet_type {
+    join = 0;
+    uplink = 1;
+  }
+
   uint64 gateway_timestamp_ms = 1;
```
madninja, line 14: rename to `tmst` to reflect that it means in packet forwarder land? Name changes are ok for already running systems

src/service/packet_router.proto
```diff
@@ -27,7 +27,7 @@ message packet_router_packet_report_v1 {
   uint32 payload_size = 11;
   bool free = 12;
   packet_type type = 13;
-  uint64 hpr_timestamp_ms = 14;
+  uint64 received_timestamp_ms = 14;
```
madninja, line 30: `received_timestamp` and add a comment above it indicating the type like mentioned in the suggestions

src/service/packet_router.proto
```diff
@@ -27,7 +27,8 @@ message packet_router_packet_report_v1 {
   uint32 payload_size = 11;
   bool free = 12;
   packet_type type = 13;
-  uint64 received_timestamp_ms = 14;
+  // Timestamp in ms
```
madninja, line 30:
```suggestion
  // Timestamp in ms since unix epoch
```

madninja (approved):

src/service/packet_router.proto
```diff
@@ -6,7 +6,12 @@ import "region.proto";
 import "data_rate.proto";
 
 message packet_router_packet_report_v1 {
-  uint64 gateway_timestamp_ms = 1;
+  enum packet_type {
+    join = 0;
+    uplink = 1;
+  }
+
+  uint64 gateway_tmst = 1;
```
jeffgrunewald, line 14: Can we safely make this change or do we need to deprecate that old field 1 and add this as a new one?

madninja: no.. it's just a name change.. same time.. and the values actually reflect the new name.. so no behavior change either

jeffgrunewald (approved):

PR #326, opened 2023-05-03
RPC for updating org records

jeffgrunewald:
Allow for updating IoT Org records in the Config Service via RPC.

The single update request message mirrors the `org_res_v1` which includes the full list of pubkeys and devaddr constraints associated to an org record via its OUI, but the `oneof` field allows for supplying a single update at a time for simplifying the API (one request to one change) while still using the separate types within the `oneof` field for the server side to switch off of when authorizing the different types of requests based on the allowed operator (administrator vs. org self-managing)

### DIFF
--- a/src/service/iot_config.proto
+++ b/src/service/iot_config.proto
@@ -152,9 +152,9 @@ message org_get_req_v1 { uint64 oui = 1; }
 
 message org_create_helium_req_v1 {
   enum helium_net_id {
-    0x00003c = 0;
-    0x60002d = 1;
-    0xc00053 = 2;
+    type0_0x00003c = 0;
+    type3_0x60002d = 1;
+    type6_0xc00053 = 2;
   }
 
   bytes owner = 1;
--- a/src/service/iot_config.proto
+++ b/src/service/iot_config.proto
@@ -152,9 +152,9 @@ message org_get_req_v1 { uint64 oui = 1; }
 
 message org_create_helium_req_v1 {
   enum helium_net_id {
-    type0_0x00003c = 0;
-    type3_0x60002d = 1;
-    type6_0xc00053 = 2;
+    0x00003c = 0;
+    0x60002d = 1;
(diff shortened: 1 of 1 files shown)


bbalser (approved):

src/service/iot_config.proto
```diff
@@ -173,6 +181,36 @@ message org_create_roamer_req_v1 {
   bytes signer = 7;
 }
 
+message org_update_req_v1 {
```
andymck, line 184: this looks odd, can you elaborate on this RPC request any why its structured with 3 sub msg types ? How does that look from the client request side of things?

jeffgrunewald: because these are all operations that update the "org" record for an iot network using organization (org record as reflected in the `org_res_v1` below)  i went back and forth on whether this should be a single update RPC or a handful of update RPCs (one for each element being updated). the `oneof` message field type convinced me to treat them as logical components of a single org update that could be executed as one-offs or by listing potential updates (the repeated list of update fields).

the rust code for the client caller would probably look something like:
```
OrgUpdateReqV1 {
    oui: 12,
    updates: vec![
        org_update_req_v1::Update::Payer(PublicKeyBinary("123DECaw532HYNJY")),
        org_update_req_v1::Update::DelegateKey(DelegateKeyUpdateV1 { delegate_key: PublicKeyBinary("1234DR564232we"), action: ActionV1::Add }),
        org_update_req_v1::Update::Devaddrs(6),
    ],
    timestamp: 1234567,
    signer: 12456abcde1234556PDERB134,
    signature: vec![], <= signed later
}
```
but on the CLI side i would _probably_ keep things simple and implement separate commands for each type of update and only construct a vec of one update operation per call. The CLI is written in rust too, so it would easily support constructing a request proto like the above but I'm guessing the actual command line input to make that happen would be more cumbersome than it's worth

macpie (approved):

PR #328, opened 2023-05-11
add gain & elevation to valid beacon and witness reports

andymck:


bbalser (approved):

jeffgrunewald (approved):

src/service/poc_lora.proto
```diff
@@ -156,6 +162,12 @@ message lora_verified_witness_report_v1 {
   // the participant to which the reason applies,
   // which rendered the report as invalid
   invalid_participant_side participant_side = 8;
+  /// the transmit gain value of the gateway in dbi x 10
+  /// For example 1 dbi = 10, 15 dbi = 150
+  /// derived from gateway metadata
+  int32 gain = 9;
+  /// The asserted elevation of the gateway, derived from gateway metadata
+  int32 elevation = 10;
```
jeffgrunewald, line 170: shouldn't this be in the `lora_valid_witness_report`? how does a "verified" witness relate to a "valid" witness?

src/service/poc_lora.proto
```diff
@@ -106,6 +106,12 @@ message lora_valid_beacon_report_v1 {
   // integer representation of a 4-point precision decimal multiplier
   // based on the number of witnesses to a poc event
   uint32 reward_unit = 5;
+  /// the transmit gain value of the gateway in dbi x 10
+  /// For example 1 dbi = 10, 15 dbi = 150
+  /// derived from gateway metadata
+  int32 gain = 6;
+  /// The asserted elevation of the gateway, derived from gateway metadata
+  int32 elevation = 7;
```
madninja, line 114: Declare units relative to what

src/service/poc_lora.proto
jeffgrunewald: confirmed the valid report for witnesses is deprecated

src/service/poc_lora.proto
andymck: unclear to me what units elevation is in, i imagine its meters but its not defined in any source proto that I am aware off.  Confirm here if you know the units

jeffgrunewald (approved):

src/service/poc_lora.proto
vihu: I'm pretty sure that elevation is AGL (above ground level) in meters.

vihu (approved):

PR #329, opened 2023-05-11
Add subscriber event

maplant:


src/service/poc_mobile.proto
```diff
@@ -4,6 +4,21 @@ package helium.poc_mobile;
 
 import "mapper.proto";
 
+message subscriber_event_req_v1 {
+  // Pub key for the subscriber's NFT
```
madninja, line 8: _helium_ pub key ideally.. I think we should try to keep all keys in oracles as close to helium keys until we sweep them all over to Solana

src/service/poc_mobile.proto
```diff
@@ -4,6 +4,21 @@ package helium.poc_mobile;
 
 import "mapper.proto";
 
+message subscriber_event_req_v1 {
```
andymck, line 7: The message name here is generic, what is we need to incorporate additional subscriber requests such as subscriber mapping requests
Suggest renaming `subscriber_location_req_v1`

src/service/poc_mobile.proto
```diff
@@ -4,6 +4,21 @@ package helium.poc_mobile;
...
+  // Pub key for the subscriber's NFT
+  bytes subscriber = 1;
```
andymck, line 9: renaming as pub_key keeps with convention of existing mobile requests

src/service/poc_mobile.proto
andymck, line 7: gonna needs a corresponding resp_v1 message

andymck:
closing as amalgamated into https://github.com/helium/proto/pull/330

PR #330, opened 2023-05-15
subscriber rewards support

andymck:
Adds message support for subscriber location sharing and subsequent rewards

### DIFF
--- a/src/service/poc_mobile.proto
+++ b/src/service/poc_mobile.proto
@@ -60,12 +60,34 @@ message mapper_attach_ingest_report_v1 {
   mapper_attach report = 2;
 }
 
+message subscriber_location_req_v1 {
+  // id of the subscriber which submitted the location report
+  bytes subscriber_id = 1;
+  // Timestamp of the subscriber location sharing event in seconds from the unix
+  // epoch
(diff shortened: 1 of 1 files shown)


src/service/poc_mobile.proto
```diff
@@ -60,12 +60,34 @@ message mapper_attach_ingest_report_v1 {
...
 }
 
+message subscriber_location_req_v1 {
+  // id of the subscriber which submitted the location report
+  bytes subscriber_id = 1;
+  // Timestamp of the subscriber location sharing event in seconds from the unix
+  // epoch
+  uint64 timestamp = 2;
+  // pubkey of the carrier identity service
+  bytes pub_key = 3;
```
jeffgrunewald, line 70: should this be `carrier_pubkey` to avoid confusion about what this pubkey refers to outside of the context of this comment above? looking at this struct in any of the oracle code i would assume it was the subscriber's helium pubkey first, followed by their solana pubkey

andymck: done

jeffgrunewald (approved):

bbalser (approved):

PR #332, opened 2023-05-18
create mapper service

lthiery:


src/mapper.proto
```diff
@@ -1,24 +1,42 @@
...
 
-message mapper {
-  oneof message {
-    mapper_gps gps = 1;
-    mapper_scan scan = 2;
-    mapper_attach attach = 3;
-  }
+service ScanUpload {
+  rpc Upload(mapper_msg) returns (scan_upload_reply) {}
+}
```
JayKickliter, line 6: it looks like this repo uses `snake_case` by convention. Is there reason this is `PascalCase`?

src/mapper.proto
```diff
@@ -1,24 +1,42 @@
...
+
+message scan_upload_reply { scan_upload_status status = 1; }
+enum scan_upload_status {
+  SUCCESS = 0;
```
JayKickliter, line 10: same comment here about casing. I would assume the code generator would do correct thing for these?

lthiery: I believe it's a convention to have then enums in `CAPITALS_WITH_UNDERSCORES`? agreed about the other one though

jeffgrunewald (commented):
what's the intention of the mapper service and where will the grpc server that `upload` rpc be running? if the messages are intended to be ingested into the iot and/or mobile oracle rewardable coverage pipelines then this service definition should probably be folded into either the `poc_mobile` or `poc_lora` services and the `upload` rpc renamed to `submit_mapper` or similar to be consistent with the other PoC data submission messages per https://github.com/helium/proto/blob/f1208e25b8966e0b480023439a53f5e159f9c0d0/src/service/poc_lora.proto#L225 and https://github.com/helium/proto/blob/f1208e25b8966e0b480023439a53f5e159f9c0d0/src/service/poc_mobile.proto#L99.

if these are meant for something else then disregard

devilish66694 (approved):

madninja:
Closed due to lack of activity

PR #333, opened 2023-05-24
subscriber id in data transfer requests and events

andymck:
Adds support for subscriber id to data pipeline

linked oracle PR: andymck/realtime-mobile-verifier-plus-location-sharing-rewards-tmp1

### DIFF
--- a/src/service/poc_mobile.proto
+++ b/src/service/poc_mobile.proto
@@ -239,6 +239,9 @@ message data_transfer_event {
   // Timestamp in seconds since the epoch
   uint64 timestamp = 7;
   bytes signature = 8;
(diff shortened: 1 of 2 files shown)


src/service/poc_mobile.proto
```diff
@@ -62,7 +62,7 @@ message mapper_attach_ingest_report_v1 {
 
 message subscriber_location_req_v1 {
   // id of the subscriber which submitted the location report
-  bytes subscriber_id = 1;
+  string subscriber_id = 1;
```
madninja, line 65: I'd prefer these to just be bytes please

bbalser (approved):

PR #334, opened 2023-05-30
Add types modules for common functionality

vihu:
Summary
----
We have noticed in multiple instances that updating/adding a proto definition would break oracles as oracles depends on gateway-rs which in turn depends on proto, thus resulting in two different versions of proto for oracles.

This PR elevates the beacon, entropy and region modules to proto itself. Subsequent PRs would:

1. [Remove gateway-rs dep from oracles](https://github.com/helium/oracles/commit/b98aef1d0e3909e62992353e6b7d3fd35bfd5e6b)
2. [Update gateway-rs to enable the types feature and use that directly](https://github.com/helium/gateway-rs/commit/b2007f201e245cb5ef57d9e737717ca3a844be00)

### Notes
- Add beacon.rs
- Add entropy.rs
- Add region.rs
- Gate behind types feature flag


lthiery:
Tested out updating [gateway-rs](https://github.com/helium/gateway-rs/commit/b2007f201e245cb5ef57d9e737717ca3a844be00) and [oracle](https://github.com/helium/oracles/commit/b98aef1d0e3909e62992353e6b7d3fd35bfd5e6b) to use this branch fairly painlessly.

lthiery:
I could imagine refactoring these types to all be under `types/beacon` and feature-gate it as `beacon-types`. But we could keep this basic architecture (ie: `types/{specific_domain}`) to extend this pattern for other common types that bubble up? Just an idea.

lthiery:
@madninja in reply to your previous comment:
> can we not have the "types" complexity. i.e. don't add new feature flags.

I think you can see in this PR that there are a lot of extra dependencies brought in by "types". ie:
```
rand = { version = "0.8", optional = true }
sha2 = { version = "0", optional = true }
base64 = { version = ">=0.21", optional = true }
byteorder = { version = "1", optional = true }
rand_chacha = { version = ">=0.3", optional = true }
rust_decimal = {version = "1", features = ["serde-with-float"], optional = true}
thiserror = { version = "1.0", optional = true }
```

That is the main reason I think a feature-gate is good. But I'm okay removing it as well if we prefer...

madninja:
> @madninja in reply to your previous comment:
> 
> > can we not have the "types" complexity. i.e. don't add new feature flags.
> 
> I think you can see in this PR that there are a lot of extra dependencies brought in by "types". ie:
> 
> ```
> rand = { version = "0.8", optional = true }
> sha2 = { version = "0", optional = true }
> base64 = { version = ">=0.21", optional = true }
> byteorder = { version = "1", optional = true }
> rand_chacha = { version = ">=0.3", optional = true }
> rust_decimal = {version = "1", features = ["serde-with-float"], optional = true}
> thiserror = { version = "1.0", optional = true }
> ```
> 
> That is the main reason I think a feature-gate is good. But I'm okay removing it as well if we prefer...

I don't think those are a big deal? They're small packages? Let's try it before we add more configuration knobs

lthiery:
> I don't think those are a big deal? They're small packages? Let's try it before we add more configuration knobs

Ok, I'll remove feature-flags in the other PR and will merge that wit this PR shortly.

lthiery:
@madninja if this all looks good, I'll open up the PRs for `gateway-rs` and `oracles`

src/lib.rs
```diff
@@ -3,8 +3,11 @@
 
 include!(concat!(env!("OUT_DIR"), "/helium.rs"));
 
+mod types;
+
 pub use blockchain_txn::Txn;
 pub use prost::{DecodeError, EncodeError, Message};
+pub use types::*;
```
madninja, line 10: can we just `pub mod types`. I've come to like the name spacing instead of exporting everything at the top level. 

The concern is that the "models" in types are not proto messages so they shouldn't live in the same namespace.

Cargo.toml
```diff
@@ -9,11 +9,18 @@ default= []
...
 bytes = "1"
 prost = "0"
+rand = "0.8"
+rand_chacha = ">=0.3"
+sha2 = "0"
+rust_decimal = { version =  "1", features = ["serde-with-float"] }
 serde = {version = "1", features = ["derive"]}
 serde_json = "1"
+tonic = {version = "0", optional = true}
+thiserror = "1.0"
```
madninja, line 23: Do we need to pin this less strict at `"1"`?

madninja (commented):
Couple of nits.. otherwise I'm ok with this.. 

The biggest oddity is that we now have a "proto" repository that has PoC specific implementation details :-/

src/lib.rs
vihu: updated [a45d99](https://github.com/helium/proto/pull/334/commits/9a45d99267e757def7782d339f7c63e78d2193bd)

Cargo.toml
vihu: check [a45d99](https://github.com/helium/proto/pull/334/commits/9a45d99267e757def7782d339f7c63e78d2193bd)

Cargo.toml
```diff
@@ -9,11 +9,18 @@ default= []
 services=["tonic-build", "tonic"]
 
 [dependencies]
-tonic = {version = "0", optional = true}
+base64 = ">=0.21"
+byteorder = "1"
 bytes = "1"
 prost = "0"
+rand = "0.8"
```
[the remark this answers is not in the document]
vihu: yeah, updated in [a45d99](https://github.com/helium/proto/pull/334/commits/9a45d99267e757def7782d339f7c63e78d2193bd)

lthiery:
> The biggest oddity is that we now have a "proto" repository that has PoC specific implementation details :-/

Definitely feels like the lesser of two evils to me. Currently, we have beacon as a sub-module to gateway-rs which depends on proto. Then oracle wanting beacon from gateway-rs but also direct acess to proto:
```
            __proto____
            |           |                     
    gateway-rs          |
    	[beacon] <----oracle		
```
As mentioned, testing updates to proto in oracle is a hassle since you have to first update gateway-rs's proto dependency and then oracles.

Now as a submodule of proto, it's more like this:
```
              proto
             __[beacon]_
            |           |                     
    gateway-rs        oracle
```

Maybe instead of feature-gating, we want to break-out types layered upon code-generated types into a separate crates within the repo instead of a sub-module of the same crate?
```
          __proto____
         |     |     |  
         | __beacon_ |
         ||         ||
     gateway-rs    oracle
```

It still improves the work-flow for updating proto from oracle, but perhaps does a better job of segregating the "POC details" from the the proto crate?

madninja:
> > The biggest oddity is that we now have a "proto" repository that has PoC specific implementation details :-/
> 
> Definitely feels like the lesser of two evils to me. Currently, we have beacon as a sub-module to gateway-rs which depends on proto. Then oracle wanting beacon from gateway-rs but also direct acess to proto:
> 
> ```
>             __proto____
>             |           |                     
>     gateway-rs          |
>     	[beacon] <----oracle		
> ```
> 
> As mentioned, testing updates to proto in oracle is a hassle since you have to first update gateway-rs's proto dependency and then oracles.
> 
> Now as a submodule of proto, it's more like this:
> 
> ```
>               proto
>              __[beacon]_
>             |           |                     
>     gateway-rs        oracle
> ```
> 
> Maybe instead of feature-gating, we want to break-out types layered upon code-generated types into a separate crates within the repo instead of a sub-module of the same crate?
> 
> ```
>           __proto____
>          |     |     |  
>          | __beacon_ |
>          ||         ||
>      gateway-rs    oracle
> ```
> 
> It still improves the work-flow for updating proto from oracle, but perhaps does a better job of segregating the "POC details" from the the proto crate?

oh the workspace approach may work out well here?

lthiery:
Yeah, I'm game for the workspace approach. I'll start by following this #57 and then add in a "beacon" crate?

madninja:
> Yeah, I'm game for the workspace approach. I'll start by following this #57 and then add in a "beacon" crate?

maybe do more like gateway-rs did since the protos themselves are the stars in this repo?

lthiery:
> maybe do more like gateway-rs did since the protos themselves are the stars in this repo?
That sounds like a plan

lthiery:
@madninja i think this setup is good. i'm going to open draft PRs on gateway-rs and oracle now

madninja (approved):

madninja:
hmm @lthiery the only niggle is that CI doesn't run tests for rust?

lthiery:
@madninja woops - thanks for catching that!

PR #335, opened 2023-06-01
beacon-type

lthiery:
Maybe this is too complicated to start, but I think this re-org reflects the beacon as the top-level module and entropy & region being sub-modules of beacon, which is how the client code seemed to see things when I was testing the change on gateway-rs and oracle.


madninja:
can we not have the "types" complexity. i.e. don't add new feature flags

lthiery:
So keep the refactor here but remove the feature flags. Note, this is a PR against #334

PR #337, opened 2023-06-08
Add ignore_empty_skf field

macpie:


michaeldjeffrey:
Hopefully this isn't too pedantic.
I'm trying to think of a field name that is almost impossible to misunderstand.
I can see depending on different expectations, someone thinking `ignore_empty_skf` means to ignore skfs and treat them as empty, rather than not routing for a devaddr with no skf.

```protobuf
bool devaddr_inactive_when_skf_empty = 8;
```

src/service/iot_config.proto
```diff
@@ -130,6 +130,8 @@ message route_v1 {
   bool active = 6;
   // Is route locked because of no payment
   bool locked = 7;
+  // Block if true when SKF empty for devaddr
```
jeffgrunewald, line 133:
```suggestion
  // If true, routes are blocked whose devaddrs have empty session key filters
```

jeffgrunewald (approved):

jeffgrunewald:
> 

i proposed a different comment to hopefully express the same; i agree with the concern but i'd prefer it as a comment to avoid typing/looking at a longer name everywhere it's used

michaeldjeffrey:
> > 
> 
> i proposed a different comment to hopefully express the same; i agree with the concern but i'd prefer it as a comment to avoid typing/looking at a longer name everywhere it's used

I've been thinking about it more, I agree I don't think it needs to be named quite as verbosely as my first comment.

But the analogy I've been using to try and think through this field is this.

---
There's a club with a bunch of rooms that all have their own admittance list.
You tell the bouncer you want to go to the Green Room. 
The list for that room is empty, and `ignore_empty_list = true`
Does the bouncer let you in?

---
The case can be made for both answers.
Maybe something like `send_for_empty_skf`? Still shorter, but more directly related to the action.

```erlang
case {Skfs, Route#route.send_for_empty_skf} of
  {[], true} -> send_packet;
  {Keys, _} -> check_keys
end
```

macpie:
> > i proposed a different comment to hopefully express the same; i agree with the concern but i'd prefer it as a comment to avoid typing/looking at a longer name everywhere it's used
> 
> I've been thinking about it more, I agree I don't think it needs to be named quite as verbosely as my first comment.
> 
> But the analogy I've been using to try and think through this field is this.
> 
> There's a club with a bunch of rooms that all have their own admittance list. You tell the bouncer you want to go to the Green Room. The list for that room is empty, and `ignore_empty_list = true` Does the bouncer let you in?
> 
> The case can be made for both answers. Maybe something like `send_when_skf_empty`? Still shorter, but more directly related to the action.
> 
> ```erlang
> case {Skfs, Route#route.send_when_skf_empty} of
>   {[], true} -> send_packet;
>   {Keys, _} -> check_keys
> end
> ```

Only issue I see with this is default should be false which would block packets and this is NOT the current behavior.

michaeldjeffrey:
> Only issue I see with this is default should be false which would block packets and this is NOT the current behavior.

`skip_when_skf_empty`?

```erlang
case {Skfs, Route#route.skip_when_skf_empty} of
  {[], false} -> send_packet;
  {Keys, _} -> check_keys
end
```

jeffgrunewald:
Did we settle on a name for the field? Protobuf bool fields default to false, but since we shouldn’t be reading old serialized message that are missing this field we can make the default either true or false; we’d have to add a new field to the database record so we can initialize all existing records with a value then and going forward enforce the desired default in either the cli or on server side or both. If we prefer the semantics of `send_when_skf_empty: true` that should be doable

PR #338, opened 2023-06-12
add protos for carrier key listing

andymck:


src/service/mobile_config.proto
```diff
@@ -156,13 +167,17 @@ service gateway {
...
 }
 
-service router {
-  // Verify a given router key with data transfer burn authority is registered
-  // returning the requested pubkey binary if present
-  rpc get(router_get_req_v1) returns (router_get_res_v1);
-  // Retrieve a list of all registered router pubkey binaries with burn
-  // authority registered to the config service
-  rpc list(router_list_req_v1) returns (router_list_res_v1);
+service subscriber {
```
andymck, line 170: it doesnt site completely right in my head to have an API where the response returns nothing more than the data specified in the request, but i get it that you are anticipating future metadata maybe coming into play for subscriber_info...

The alternative at this point is to have an API to determine if a specified subscriber id is known, server returns a bool.

Either works for the current needs of the mobile verifier

PR #340, opened 2023-06-13
convert router to carrier key naming; replace key type with role

jeffgrunewald:
the bulk of the "carrier" (formerly router) key validation functionality was already implemented on the `router` service and its associated messages, but we didn't have a great definition for the details of the feature at the time. b/c of that, i believe it's pretty safe to update these RPCs/messages as they haven't been utilized as of yet in the production mobile config service and we can just replace the old definitions in-place

### DIFF
--- a/src/service/mobile_config.proto
+++ b/src/service/mobile_config.proto
@@ -65,8 +65,8 @@ message gateway_info_stream_res_v1 {
   bytes signature = 4;
 }
 
-message router_get_req_v1 {
-  // Pubkey binary of the requested router
+message carrier_get_req_v1 {
+  // Pubkey binary of the requested carrier/router
   bytes pubkey = 1;
   // pubkey binary of the signing keypair
   bytes signer = 2;
@@ -74,8 +74,8 @@ message router_get_req_v1 {
   bytes signature = 3;
 }
 
-message router_get_res_v1 {
-  // Pubkey binary of the requested router, confirming registration
+message carrier_get_res_v1 {
+  // Pubkey binary of the requested carrier/router, confirming registration
   bytes pubkey = 1;
   // unix epoch timestamp in seconds
   uint64 timestamp = 2;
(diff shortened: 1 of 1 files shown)


src/service/mobile_config.proto
```diff
@@ -104,17 +111,19 @@ message router_list_res_v1 {
 }
 
 message admin_add_key_req_v1 {
```
andymck, line 113: when adding a key the `key_role` is required to be supplied, can one key be added to more than one role ?

I note when removing a key, no `key_role` is required in the request....should it be ?

src/service/mobile_config.proto
```diff
@@ -65,17 +65,17 @@ message gateway_info_stream_res_v1 {
...
   // pubkey binary of the signing keypair
   bytes signer = 2;
   // Signature over the request by the requesting oracle
   bytes signature = 3;
 }
 
-message router_get_res_v1 {
-  // Pubkey binary of the requested router, confirming registration
+message carrier_get_res_v1 {
+  // Pubkey binary of the requested carrier/router, confirming registration
```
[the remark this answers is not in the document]
jeffgrunewald: removed; not needed or used

src/service/mobile_config.proto
```diff
@@ -65,17 +65,17 @@ message gateway_info_stream_res_v1 {
   bytes signature = 4;
 }
 
-message router_get_req_v1 {
-  // Pubkey binary of the requested router
+message carrier_get_req_v1 {
```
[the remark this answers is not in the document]
jeffgrunewald: removed

src/service/mobile_config.proto
```diff
@@ -65,36 +65,34 @@ message gateway_info_stream_res_v1 {
...
-  bytes signature = 2;
+  bytes signature = 3;
 }
 
-message router_list_res_v1 {
-  // List of public key binaries of all registered routers
-  repeated bytes routers = 1;
+message authorization_list_res_v1 {
+  // List of public key binaries of all registered entities by requested role
+  repeated bytes entity = 1;
```
andymck, line 95: why use entity here when elsewhere pubkey is in use, ie when a key is added its called a pubkey but when listed its an entity

jeffgrunewald: fixed [c71ed9c](https://github.com/helium/proto/pull/340/commits/c71ed9c16fc1efdff1d2aed1294f1f47dd93c8bc)

bbalser (approved):

andymck (approved):

PR #344, opened 2023-06-20
Modeled coverage

maplant:


src/service/poc_mobile.proto
```diff
@@ -80,6 +83,41 @@ message subscriber_location_ingest_report_v1 {
...
+enum signal_level {
+  NO = 0;
+  LOW = 1;
+  MEDIUM = 2;
+  HIGH = 3;
+}
+
+// Radio signal level and power in the h3 hex.
+message radio_hex_signal_level {
+  uint64 hex_id = 1;
```
madninja, line 95: what is `hex_id` an h3 hex? if so that's called `location`?

src/service/poc_mobile.proto
```diff
@@ -80,6 +83,41 @@ message subscriber_location_ingest_report_v1 {
...
+  signal_level signal_level = 2;
+  double signal_power = 3;
```
madninja, line 97: comment with type info?

src/service/poc_mobile.proto
maplant: It is an h3 hex. I see other places use h3 hexes via a string called location, so I will change to that

src/service/poc_mobile.proto
maplant: dBm, apparently

madninja: Can we use a u32 or something like we do with IoT PoC and use ddbm? trying to avoid floats since they're massive and terrible language support

src/service/poc_mobile.proto
```diff
@@ -80,6 +83,43 @@ message subscriber_location_ingest_report_v1 {
...
+enum signal_level {
+  NO = 0;
+  LOW = 1;
+  MEDIUM = 2;
+  HIGH = 3;
+}
+
+// Radio signal level and power in the h3 hex.
+message radio_hex_signal_level {
+  // H3 hex of the location of the radio
```
kurotych, line 95: About comment. 

It may be a hex location covered by radio, but not a location of the radio itself. 
Could you confirm? @thehardbits

maplant: ah, right, that's an important distinction. Thanks guys!

src/service/poc_mobile.proto
```diff
@@ -80,6 +83,43 @@ message subscriber_location_ingest_report_v1 {
...
+  double signal_power = 3;
+}
+
+// Array of radio_hex_signal_level messages
+message signal_levels { repeated radio_hex_signal_level value = 1; }
+
+message coverage_object_req_v1 {
+  bytes uuid = 1;
+  string cbsd_id = 2;
+  uint64 coverage_claim_time = 3;
```
kurotych, line 108: Could you add a comment about the measurement of `coverage_claim_time`?
ms or sec?

maplant: Sure. I'm assuming it's a timestamp in seconds? Or ms?

kurotych: seconds

src/service/mobile_config.proto
```diff
@@ -95,6 +95,7 @@ enum admin_key_role {
 enum network_key_role {
   mobile_carrier = 0;
   mobile_router = 1;
+  mobile_pcs = 2;
```
jeffgrunewald, line 98: maybe comment these so they have explanations what they mean? this will also need a complimentary migration in the oracles to add a `pcs` to the database key_role enum.

src/service/poc_mobile.proto
```diff
@@ -80,6 +83,42 @@ message subscriber_location_ingest_report_v1 {
...
+  NO = 0;
+  LOW = 1;
+  MEDIUM = 2;
+  HIGH = 3;
+}
+
+// Radio signal level and power in the h3 hex.
+message radio_hex_signal_level {
+  // H3 hex tile covered by the radio
+  string location = 1;
```
jeffgrunewald, line 96: does this have a constant hex resolution we can reference here as well, as in "H3 res12 hex tile"?

src/service/poc_mobile.proto
```diff
@@ -80,6 +83,42 @@ message subscriber_location_ingest_report_v1 {
...
+  signal_level signal_level = 2;
+  // Signal power of the radio in dBm (decibel-milliwatts)
+  double signal_power = 3;
```
jeffgrunewald, line 99: i see a resolved comment but no explanation? don't we typically have this type of field as `uint32` or `uint64` with the ddBm scale comment?

src/service/poc_mobile.proto
```diff
@@ -80,6 +83,42 @@ message subscriber_location_ingest_report_v1 {
...
+}
+
+message coverage_object_req_v1 {
+  bytes pub_key = 1;
+  bytes uuid = 2;
```
jeffgrunewald, line 104: where does this `uuid` come from? how does it differ from the `pub_key` and the `cbsd_id`?

src/service/poc_mobile.proto
```diff
@@ -99,13 +140,31 @@ message file_info {
...
+  repeated radio_hex_signal_level coverage = 5;
+  bool indoor = 6;
+  coverage_object_validity validity = 7;
+}
+
+enum coverage_object_validity {
+  coverage_object_validity_valid = 0;
+  coverage_object_validity_invalid_pub_key = 1;
+  coverage_object_validity_indoor_coverage_is_medium = 2;
+}
```
jeffgrunewald, line 157: this seems largely redundant with the `coverage_object_req_v1` message type and i don't see the `coverage_object` message used anywhere. it also looks a bit confusing with the `coverage_object` message type and the `coverage_object` field name referring to a UUID in a couple places.

jeffgrunewald: i see the resolved comments requesting to change `coverage_object_uuid` to `coverage_object` but it's still confusing seeing that field name in conjunction with a message type of the same name. if `coverage_object` message _is_ still required i would change the `coverage_object` field name to `object_id` or something. i disagree that `coverage_object` implies a reference when there's a message with that name; it implies to me the object message

bbalser (approved):

jeffgrunewald (approved):

PR #348, opened 2023-06-27
fix dup enum values

andymck:


bbalser (approved):

src/service/poc_mobile.proto
```diff
@@ -275,10 +275,10 @@ message data_transfer_session_ingest_report_v1 {
...
 
 enum data_transfer_ingest_report_status {
-  valid = 0;
-  cancelled = 1;
-  invalid_gateway_key = 2;
-  invalid_routing_key = 3;
+  dt_valid = 0;
+  dt_cancelled = 1;
+  dt_invalid_gateway_key = 2;
+  dt_invalid_routing_key = 3;
```
jeffgrunewald, line 281: alternatively, if these are _only_ used in the `invalid_data_transfer_ingest_report` message you could nest this enum inside that message and not need the prefix.
```
message invalid_data_transfer_ingest_report_v1 {
  enum report_status {
    valid = 0;
    cancelled = 1;
    invalid_gateway_key = 2;
    invalid_routing_key = 3;
  }

  data_transfer_session_ingest_report_v1 report = 1;
  report_status reason = 2;
  uint64 timestamp = 3;
```

andymck: k that works, good to know this one

PR #349, opened 2023-06-29
Mapper types

lthiery:
Provides handwritten Rust types for Mapper messages so as to provide more specific types for the fields.

Also includes lora payload ser/deser for Attach and ScanBeacon.

### DIFF
--- a/mapper/Cargo.toml
+++ b/mapper/Cargo.toml
@@ -1,0 +1,21 @@
+[package]
+name = "mapper_messages"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+chrono = { version = "0", features = ["serde"] }
+geo-types = "0"
+hex = "0"
+helium-proto = { path = "../", features = ["services"] }
+h3o = "0"
+lazy_static = "1"
+nmea-parser = "0"
+rust_decimal = "1"
+serde =  {version = "1", features = ["derive"] }
+serde_json = "1"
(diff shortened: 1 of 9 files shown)


mapper/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "mapper_messages"
```
vihu, line 2: Just call this mapper? Unless there's a namespace conflict?

mapper/src/cell_attach.rs
```diff
@@ -0,0 +1,298 @@
+use super::gps::{altitude, hdop, latlon, speed, time};
+use super::*;
+
+#[derive(Serialize, Debug, Clone, PartialEq)]
+pub struct CellAttach {
+    // This allows us to detect censorship efforts. It can roll over.
```
vihu, line 6: What exactly does roll over mean for attach_counter?

mapper/Cargo.toml
lthiery: Yes, I think there will be a namespace conflict when the application on spot-firmware called mapper reaches for it.

mapper/src/cell_attach.rs
lthiery: Each attach attempt gets counted. We probably won't ever do more than 4.3M attaches, but you never know :shrug:

PR #359, opened 2023-07-20
Session Key

macpie:


src/service/packet_router.proto
```diff
@@ -52,10 +52,38 @@ message packet_router_register_v1 {
   bytes signature = 3;
 }
 
+// Session offer from the packe router to the gateway. If the gateway wants to
```
mawdegroot, line 55:
```suggestion
// Session offer from the packet router to the gateway. If the gateway wants to
```

src/service/packet_router.proto
```diff
@@ -42,7 +42,7 @@ message packet_router_packet_up_v1 {
   float snr = 6;
   region region = 7;
   uint64 hold_time = 8;
-  bytes gateway = 9;
+  bytes session_key = 9;
```
mawdegroot, line 45: is this right?

madninja: Fixed in [85ac774](https://github.com/helium/proto/pull/359/commits/85ac77448b3d26c74a259a80583ec0f1d92a0b49)

src/service/packet_router.proto
```diff
@@ -52,10 +52,38 @@ message packet_router_register_v1 {
   bytes signature = 3;
 }
 
+// Session offer from the packet router to the gateway. If the gateway wants to
```
macpie, line 55: I think `bytes gateway = 9;` in `packet_router_packet_up_v1` above should signal that it is not always the gateway but the public session key, no?

src/service/packet_router.proto
```diff
@@ -52,10 +52,38 @@ message packet_router_register_v1 {
...
+//
+// This message is sent by packet router after the register command is sent by
+// the gateway.
+message packet_router_session_offer_v1 { bytes data = 1; }
+
+// Initializes a session key with the packet router. The data in this init
+// message has to be from the latest session offer the gateway has received, the
+// session key is the public key that is to be used to verify packets by packet
+// router, and this message is to be signed by the private key of the gateway.
+// Once accepted the session key replaces the current (default is gateway) key
```
macpie, line 67: I think bytes `gateway = 9;` in `packet_router_packet_up_v1` above should signal that it is not always the gateway but the public session key also, maybe we can come up with a middle ground?

madninja:
No because it’s always the gateway key. The session key is _not_ supplied in the packet

src/service/packet_router.proto
madninja: I don't agree here.. The session key is a protocol optimization between gateways and HPR.. By keeping the gateway as is you're only overriding the signature if an alternate verifier was set as part of session negotiation

macpie: OK, I see. So HPR should now about session (obviously) and check signature based on that knowledge only?

madninja: yeah.. basically the "verifier" used in the packet router is set to the gateway public key as part of registration (or before?) and the session key becomes the verifier.. you could technically ignore the gateway bytes in the packet report altogether in that model

src/service/packet_router.proto
```diff
@@ -52,10 +52,38 @@ message packet_router_register_v1 {
   bytes signature = 3;
 }
 
+// Session offer from the packet router to the gateway. If the gateway wants to
+// initiate a session key it should send a packet_router_session_init_v1 message
+// to the packet router using data in this offer to propose a session key.
+//
+// This message is sent by packet router after the register command is sent by
+// the gateway.
+message packet_router_session_offer_v1 { bytes data = 1; }
```
ke6jjj, line 61: Can this please be called "nonce" instead of "data". That more accurately describes what this field does, and since we're super sticklers about never renaming a protobuf field after its deployment, this is our last chance to change it.

src/service/packet_router.proto
```diff
@@ -52,10 +52,38 @@ message packet_router_register_v1 {
...
+// as the verifier used by the packet router for upilnk packets.
+//
+// The session key is valid for the length of the current stream or until a next
+// offer message is received from the packet router.
+//
+// Once the new verifier is set by the packet router, any signed packets that do
+// not pass verification with the active verifier key are dropped.
+message packet_router_session_init_v1 {
+  bytes gateway = 1;
+  bytes data = 2;
```
ke6jjj, line 77: "nonce" here as well.

src/service/packet_router.proto
madninja: Fixed in [1e49483](https://github.com/helium/proto/pull/359/commits/1e4948358077027813cd81b70b30d20e67aa5901)

src/service/packet_router.proto
madninja: Fixed in [1e49483](https://github.com/helium/proto/pull/359/commits/1e4948358077027813cd81b70b30d20e67aa5901)

src/service/packet_router.proto
ke6jjj: ty!

ke6jjj (approved):

ke6jjj (approved):

PR #364, opened 2023-08-07
Add wifi_heartbeat_req_v1

bbalser:
Add new message to receive heartbeats from wifi hotspots

Linked oracle pr: https://github.com/helium/oracles/pull/623

### DIFF
--- a/src/service/poc_mobile.proto
+++ b/src/service/poc_mobile.proto
@@ -217,6 +217,9 @@ message heartbeat {
   // Timestamp of location validation in seconds from unix epoch
   // applicable to wifi heartbeats only
   uint64 location_validation_timestamp = 10;
+  // Distance in meters to the asserted location of the gateway_reward
+  // at the time of heartbeat verification
+  uint64 distance_to_asserted = 11;
 }
 
 enum heartbeat_validity {
--- a/src/service/poc_mobile.proto
+++ b/src/service/poc_mobile.proto
@@ -68,6 +68,28 @@ message cell_heartbeat_ingest_report_v1 {
   cell_heartbeat_req_v1 report = 2;
 }
 
+message wifi_heartbeat_req_v1 {
+  // Public key of the hotspot
+  bytes pub_key = 1;
+  // Timestamp of heartbeat in seconds from unix epoch
+  uint64 timestamp = 2;
+  double lat = 3;
+  double lon = 4;
+  // Timestamp of location validation in seconds from unix epoch
+  uint64 location_validation_timestamp = 5;
+  bool operation_mode = 6;
+  bytes coverage_object = 7;
+  bytes signature = 8;
+}
+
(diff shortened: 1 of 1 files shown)


src/service/poc_mobile.proto
```diff
@@ -56,6 +56,28 @@ message cell_heartbeat_ingest_report_v1 {
...
+  // Public key of the hotspot
+  bytes pub_key = 1;
+  // Timestamp of heartbeat in seconds from unix epoch
+  uint64 timestamp = 2;
+  double lat = 3;
+  double lon = 4;
+  // Timestamp of location validation in seconds from unix epoch
+  uint64 location_validation_timestamp = 5;
+  bool operation_mode = 6;
+  bytes coverage_object = 7;
```
andymck, line 69: why is this bytes  and not a coverage_object type ?

bbalser: this is the uuid of the coverage_object in bytes just like in the cell_heartbeat https://github.com/helium/proto/blob/302880a665f92db5ac865a1878444ec87d1a80e6/src/service/poc_mobile.proto#L47

andymck: k that makes sense then

src/service/poc_mobile.proto
```diff
@@ -56,6 +56,28 @@ message cell_heartbeat_ingest_report_v1 {
   cell_heartbeat_req_v1 report = 2;
 }
 
+message wifi_heartbeat_req_v1 {
+  // Public key of the hotspot
+  bytes pub_key = 1;
+  // Timestamp of heartbeat in seconds from unix epoch
+  uint64 timestamp = 2;
+  double lat = 3;
```
andymck, line 64: using double here for lat and lon contravenes the protobuf guidelines defined in the repo readme.

We use the same in other proto messages so maybe its an exception or maybe it flew in under the radar

madninja: lat/lon has been exempted because it was used in mobile heartbeats.. It'll have to be used in poc/reward activity calculations _very_ carefully given architecture differences

src/service/poc_mobile.proto
```diff
@@ -308,6 +336,8 @@ enum cell_type {
   neutrino430 = 2;
   sercomm_indoor = 3;
   sercomm_outdoor = 4;
+  cell_type_none = 5;
+  nova_generic_wifi_indoor = 6;
```
jeffgrunewald, line 340: do we need `generic`? is missing that distinction a meaningful thing for the `secomm_indoor/outdoor` cbrs radios above?

jeffgrunewald (approved):

src/service/poc_mobile.proto
andymck: generic was agreed so that we can permit additional future models to be added

maplant (approved):

PR #366, opened 2023-08-17
support invalid details for invalid beacons and witnesses

andymck:
Adds an invalid_details field which can be populated with context on why a POC report was declared invalid.  The invalid_details value should be descriptive and add additional value to the denied reason.

### DIFF
--- a/src/service/poc_lora.proto
+++ b/src/service/poc_lora.proto
@@ -160,6 +160,9 @@ message lora_invalid_beacon_report_v1 {
   /// The asserted elevation of the gateway in AGL ( above ground level)
   /// derived from gateway metadata
   int32 elevation = 6;
+  // provides any additional context for invalid reason
+  // for example the deny list version used as part of the deny list check
+  string invalid_details = 7;
 }
 
 // tagged invalid witness report produced by the verifier
@@ -171,6 +174,9 @@ message lora_invalid_witness_report_v1 {
   // the participant to which the reason applies,
   // which rendered the report as invalid
   invalid_participant_side participant_side = 4;
+  // provides any additional context for invalid reason
+  // for example the deny list version used as part of the deny list check
+  string invalid_details = 5;
 }
 
 // tagged verified witness report produced by the verifier
@@ -196,6 +202,9 @@ message lora_verified_witness_report_v1 {
   /// The asserted elevation of the gateway in AGL ( above ground level)
   /// derived from gateway metadata
   int32 elevation = 10;
+  // provides any additional context for invalid reason
+  // for example the deny list version used as part of the deny list check
+  string invalid_details = 11;
 }
 
 // POC report produced by the verifier
--- a/src/service/poc_lora.proto
+++ b/src/service/poc_lora.proto
@@ -145,6 +145,10 @@ message lora_valid_witness_report_v1 {
   uint32 reward_unit = 5;
 }
 
+message invalid_details {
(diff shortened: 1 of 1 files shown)


bbalser (approved):

jeffgrunewald (approved):

src/service/poc_lora.proto
```diff
@@ -145,6 +145,10 @@ message lora_valid_witness_report_v1 {
   uint32 reward_unit = 5;
 }
 
+message invalid_details {
+  oneof data { string denied = 1; }
```
jeffgrunewald, line 149: if this is now a oneof, should the denylist version be an integer instead of a string since the format is 20230817?

bbalser: I vote leave it as a string since the source is a github tag, which could be something different in the future even if that is unlikely.

andymck: yeah string is more change of being future proof

bbalser (approved):

src/service/poc_lora.proto
madninja: agreed.. just change it to denylsit_version perhaps?

src/service/poc_lora.proto
```diff
@@ -160,6 +164,9 @@ message lora_invalid_beacon_report_v1 {
   /// The asserted elevation of the gateway in AGL ( above ground level)
   /// derived from gateway metadata
   int32 elevation = 6;
+  // provides any additional context for invalid reason
+  // for example the deny list version used as part of the deny list check
+  invalid_details invalid_details = 7;
```
madninja, line 169: rename invalid_details to `details`.. i.e. `invalid_details details = 7;` to match the invalid_reason setup

madninja (commented):
Couple of nits around naming otherwise g2g

src/service/poc_lora.proto
andymck: i went with denylist_tag, which is more contextually correct

src/service/poc_lora.proto
andymck: kept it as is, the reason field has two naming conventions depending on msg type, it will be one of "reason" or "invalid_reason"....ideally they would have been consistent but alas here we are
Rather than go with two naming conventions for the details field, i am using `invalid_details` naming in all types. This works better in the oracle side, allowing for a consistent naming convention to be utilised

src/service/poc_lora.proto
```diff
@@ -171,6 +178,9 @@ message lora_invalid_witness_report_v1 {
   // the participant to which the reason applies,
   // which rendered the report as invalid
   invalid_participant_side participant_side = 4;
+  // provides any additional context for invalid reason
+  // for example the deny list version used as part of the deny list check
+  invalid_details invalid_details = 5;
```
madninja, line 183: rename this one too!

src/service/poc_lora.proto
```diff
@@ -196,6 +206,9 @@ message lora_verified_witness_report_v1 {
   /// The asserted elevation of the gateway in AGL ( above ground level)
   /// derived from gateway metadata
   int32 elevation = 10;
+  // provides any additional context for invalid reason
+  // for example the deny list version used as part of the deny list check
+  invalid_details invalid_details = 11;
```
madninja, line 211: and this one!

madninja (approved):

madninja (approved):

PR #367, opened 2023-08-28
Add seniority timestamp and coverage object to radio reward output

maplant:


src/service/poc_mobile.proto
```diff
@@ -260,6 +260,10 @@ message radio_reward {
   uint64 poc_reward = 4;
   // Accumulated coverage points for the radio
   uint64 coverage_points = 5;
+  // Seniority timestamp for this radio
```
bbalser, line 263: comment should say if timestamp is in seconds or millis

src/service/poc_mobile.proto
```diff
@@ -260,6 +260,10 @@ message radio_reward {
...
+  uint64 seniority_timestamp = 6;
+  // Coverage object used to reward this radio
```
bbalser, line 265: comment should say this is the coverage_object uuid

bbalser (approved):

PR #377, opened 2023-11-14
Andymck/service provider rewards

andymck:
Service provider rewards
Also includes provision for unallocated rewards

### DIFF
--- a/src/service/mobile_config.proto
+++ b/src/service/mobile_config.proto
@@ -93,6 +93,24 @@ message entity_verify_res_v1 {
   bytes signature = 4;
 }
 
+message carrier_key_to_entity_req_v1 {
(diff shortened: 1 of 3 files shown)


src/service/mobile_config.proto
```diff
@@ -81,6 +81,23 @@ message entity_verify_res_v1 {
   bytes signature = 4;
 }
 
+message carrier_key_to_entity_req_v1 {
+  // helium pubkey binary of the carrier
+  string pubkey = 1;
```
jeffgrunewald, line 86: should be bytes if a pubkeybinary

andymck: comment updated, intent is to have string representation

PR #378, opened 2023-11-16
Add fields to heartbeat and mobile_reward_share files to help understand how rewards are calculated

bbalser:
In troubleshooting various issues, I realized how valuable having all the multipliers be recorded in the mobile_verifier outputs would be.

### DIFF
--- a/src/service/poc_mobile.proto
+++ b/src/service/poc_mobile.proto
@@ -214,7 +214,7 @@ message heartbeat {
   string cbsd_id = 1;
   bytes pub_key = 2;
   // this value is the heartbeat_multiplier * location_trust_score_multiplier
-  float reward_multiplier = 3;
+  float reward_multiplier = 3 [ deprecated = true ];
   uint64 timestamp = 4;
   cell_type cell_type = 5;
   heartbeat_validity validity = 6;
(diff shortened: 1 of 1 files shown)


madninja:
> 2. Fair point, should we just make version 2 of these messages to avoid that behaviour or is there something else you had in mind?

Not sure that'll help since you would have to decide what to do with the v1 protos anyway? Are the default `0` not used otherwise? If they're not "normal" values that may be enough to tell the two apart?

src/service/poc_mobile.proto
```diff
@@ -321,13 +325,24 @@ message radio_reward {
...
 }
 
 message gateway_reward {
   /// Public key of the hotspot
   bytes hotspot_key = 1;
   /// Amount awarded for dc transfer
   uint64 dc_transfer_reward = 2;
+  /// count of rewardable bytes transfered
+  uint64 rewardable_bytes = 3;
+  /// Price of MOBILE used when calculating rewards
```
andymck, line 344:
```suggestion
  /// Price of MOBILE @ 10^6 used when calculating rewards
```

madninja (approved):

riobah:
1. Is there a planned release date for these changes? 
2. the comments in code imply some equation like`reward_multiplier = heartbeat_multiplier * location_trust_score_multiplier`. In which message is this `heartbeat_multiplier`?
3. when `reward_multiplier` is deprecated, does it mean that it will not be maintained and come as 0 immediately? 
4. To give full context of the reward calculations, do we need some heartbeat multiplier in the `radio_reward` message?

PR #379, opened 2023-11-17
Andymck/verify hb cell type

andymck:


src/service/mobile_config.proto
```diff
@@ -21,15 +21,15 @@ message gateway_metadata {
...
-  string device_type = 3;
 }
 
 message gateway_info {
   // The public key binary address and on-chain identity of the gateway
   bytes address = 1;
   // The gateway metadata as recorded on the blockchain
   gateway_metadata metadata = 2;
+  // the asserted device type of the gateway
+  device_type device_type = 3;
```
madninja, line 32: why did you move this out of gateway_metadata?

andymck: Internally metadata is an Option which makes accessing fields a lil more cumbersome but you are right in that the device_type is onchain as part of the metadata. I can move it back, no big deal

bbalser (approved):

src/service/mobile_config.proto
jeffgrunewald: After discussing we decided to revert/keep the device type in the top-level GatewayInfo. Key reasons were:
- it's similar to `is_full_hotspot` in that it should be an intrinsic aspect of the hotspot instead of a fungible element like asserted location
- it makes the handling of the data easier with fewer nested "options"

PR #384, opened 2024-01-05
iot unallocated rewards

andymck:


src/service/poc_lora.proto
```diff
@@ -237,6 +237,21 @@ message operational_reward {
...
+  unallocated_reward_type_oracle = 2;
+  unallocated_reward_type_data = 3;
+}
+
+message unallocated_reward {
+  // the reward type representing a reward category to which an unallocated
+  // amount exists
+  unallocated_reward_type reward_type = 1;
+  /// Amount in bones unallocated
+  uint64 amount = 2;
```
jeffgrunewald, line 252: these are always hnt bones? no need to denote the token type of the unallocated amount?

andymck: all reward outputs are in bones, yes.  I am not sure what you are asking on the second question

jeffgrunewald: asking specifically if this is _HNT_ bones, _IoT_ bones, or if it could be either or ever

andymck: gotcha now, these are iot bones and only iot bones...i'll update the comment to include 'iot'

andymck: fixed here 96d71c04056ddc741f5e89aef6ce34fe3ce4bb80

jeffgrunewald (approved):

PR #389, opened 2024-01-17
hex boosting protos

andymck:
oracle PR: https://github.com/helium/oracles/pull/716/files

### DIFF
--- a/src/hex_boosting.proto
+++ b/src/hex_boosting.proto
@@ -1,0 +1,31 @@
+syntax = "proto3";
+
+package helium;
+
+message boosted_hex_info_v1 {
+  // The res12 h3 index of the boosted hex
+  uint64 location = 1;
+  // Unix timestamp in seconds of the start of the boost period
+  // 0 if the boosting has not yet started
+  uint64 start_ts = 2;
+  // Unix timestamp in seconds of the end of the boost period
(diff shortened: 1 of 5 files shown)


src/hex_boosting.proto
```diff
@@ -17,10 +17,10 @@ message boosted_hex_info_v1 {
   // the multipliers valid for this hex
   // for each period
   repeated uint32 multipliers = 5;
-  // the onchain pubkey of the boosted hex account
-  bytes boosted_hex_pubkey = 6;
-  // the onchain pubkey of the boost config account
-  bytes boost_config_pubkey = 7;
+  // the onchain address of the boosted hex account
+  string boosted_hex_pubkey = 6;
```
madninja, line 21: these need to be bytes..

andymck: why so, on chain they are varchar?

andymck: the varchar contains stringified version of the pubkey ie `BZM1QTud72B2cpTW7PhEnFmRX7ZWzvY7DpPpNJJuDrWG`

i can put them back to bytes in the proto, i have no requirements to have them in the proto as string...i was just keeping them in the original format...

bbalser (approved):

PR #392, opened 2024-02-26
Add oracle boosting report

maplant:
Unresolved questions:
- Do we attach the filename of the new datasets to the message?


bbalser:
Should we add some identifier of the radio/hotspot?

bbalser (approved):

PR #401, opened 2024-04-26
Add `boosted_hex_device_type_v1` for HIP-109

michaeldjeffrey:
The first iteration of HIP-109[1] adds the ability to boost by a device type. Pre-existing boosts will continue to boost _all_ device types.

From the way the HIP reads, at some point, a hex may be boosted by the same provider for multiple device types. It was undecided wether that would be implemented as multiple device_types attached to a boosted hex, or multiple instances of a boosted hex.

[1]: https://github.com/helium/HIP/blob/main/0109-hex-boosting-by-deployment.md

### DIFF
--- a/src/hex_boosting.proto
+++ b/src/hex_boosting.proto
@@ -23,6 +23,8 @@ message boosted_hex_info_v1 {
   bytes boost_config_pubkey = 7;
 
   uint32 version = 8;
+  // The device type(s) this boost applies to.
+  boosted_hex_device_type_v1 device_type = 9;
 }
 
 message boosted_hex_update_v1 {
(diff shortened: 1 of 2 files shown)


src/hex_boosting.proto
```diff
@@ -31,3 +33,11 @@ message boosted_hex_update_v1 {
   // Details of the updated hex
   boosted_hex_info_v1 update = 2;
 }
+
+enum boosted_hex_device_type_v1 {
```
andymck, line 37: nit: can this enum be moved to the top of the file.  Things read better when sub types are defined prior to their use

devilish66694 (approved):

PR #409, opened 2024-07-09
Add verified_subscriber_mapping_event_v1

macpie:


riobah:
I'm aware it can be a bit early for such questions, but will we have data about mapping events per hotspots? 
some message when a hotspot is verified by a mapper, holding fields like `pub_key`, `timestamp`, `subscriber_id` etc?

I tried to check if we have any existing data for that purpose, but I believe there is no such data made available yet.

src/service/poc_mobile.proto
```diff
@@ -664,3 +666,27 @@ verified_service_provider_boosted_rewards_banned_radio_ingest_report_v1 {
...
+
+message subscriber_mapping_event_req_v1 {
+  /// The id of the discovery mapping enabled subscriber
+  bytes subscriber_id = 1;
+  /// The accumulated mapping points the subscriber has earned this epoch
+  uint64 total_reward_points = 2;
+  /// Unix timestamp in seconds of when the mapping event was generated
+  uint64 timestamp = 3;
+  /// Pubkey of verification mapping event service
+  bytes pub_key = 6;
```
andymck, line 678: maybe rename pub_key to verification_mapping_pubkey

src/service/poc_mobile.proto
```diff
@@ -664,3 +666,27 @@ verified_service_provider_boosted_rewards_banned_radio_ingest_report_v1 {
   // epoch
   uint64 timestamp = 3;
 }
+
+message subscriber_mapping_event_req_v1 {
```
andymck, line 670: the number is off in this msg, jumps from 3 to 6

src/service/poc_mobile.proto
```diff
@@ -664,3 +666,27 @@ verified_service_provider_boosted_rewards_banned_radio_ingest_report_v1 {
...
+  uint64 timestamp = 3;
+  /// Pubkey of verification mapping event service
+  bytes pub_key = 6;
+  /// Signed payload of the verification mapping event service
+  bytes signature = 7;
+}
+
+message subscriber_mapping_event_res_v1 { string id = 1; }
+
+message verified_subscriber_mapping_event_v1 {
```
andymck, line 685: i dont understand what this msg is for ? You should have another msg that is outputted from the ingestor which wraps the original subscriber_mapping_event_req_v1 msg...see my original proto branch here for reference: https://github.com/helium/proto/blob/andymck/subscriber-mapping-support/src/service/poc_mobile.proto#L228

andymck:
> I'm aware it can be a bit early for such questions, but will we have data about mapping events per hotspots? some message when a hotspot is verified by a mapper, holding fields like `pub_key`, `timestamp`, `subscriber_id` etc?
> 
> I tried to check if we have any existing data for that purpose, but I believe there is no such data made available yet.

I dont believe there are plans to make that data public as it would be a security risk

src/service/poc_mobile.proto
macpie: ok I made it more like your branch

src/service/poc_mobile.proto
```diff
@@ -664,3 +667,24 @@ verified_service_provider_boosted_rewards_banned_radio_ingest_report_v1 {
...
+
+message verified_subscriber_mapping_event_req_v1 {
+  /// The id of the discovery mapping enabled subscriber
+  bytes subscriber_id = 1;
+  /// The accumulated mapping points the subscriber has earned this epoch
+  uint64 total_reward_points = 2;
+  /// Unix timestamp in seconds of when the mapping event was generated
+  uint64 timestamp = 3;
+  /// Pubkey of verification mapping event service
+  bytes verification_mapping_pubkey = 4;
```
andymck, line 679: this should be changed back to carrier_mapping_key in order to keep with existing conventions.  The key naming is based on the type of key rather than the name of the service and in this case the key is of type carrier

src/service/poc_mobile.proto
```diff
@@ -664,3 +667,24 @@ verified_service_provider_boosted_rewards_banned_radio_ingest_report_v1 {
   // epoch
   uint64 timestamp = 3;
 }
+
+message verified_subscriber_mapping_event_req_v1 {
+  /// The id of the discovery mapping enabled subscriber
+  bytes subscriber_id = 1;
+  /// The accumulated mapping points the subscriber has earned this epoch
+  uint64 total_reward_points = 2;
+  /// Unix timestamp in seconds of when the mapping event was generated
```
andymck, line 676: we should be using millisec precision, so lets update this comment

PR #413, opened 2024-08-19
Add subscriber referral eligibility ingest messages

maplant:


src/service/poc_mobile.proto
```diff
@@ -278,6 +278,33 @@ message verified_subscriber_location_ingest_report_v1 {
...
+enum subscriber_referral_eligibility_status {
+  referral_eligibility_status_valid = 0;
+  referral_eligibility_status_invalid_subscriber_id = 1;
+  referral_eligibility_status_invalid_carrier_key = 2;
+}
+
+message subscriber_referral_eligibility_ingest_report_v1 {
+  subscriber_referral_eligibility_req_v1 report = 1;
+  // Status of the report
+  subscriber_referral_eligibility_status status = 2;
```
bbalser, line 303: Not sure why the ingest report has a status.  I believe you need to create a `verified_subscriber_referral_eligibility_ingest_report_v1` and the status should be there.

src/service/poc_mobile.proto
```diff
@@ -566,6 +609,20 @@ message subscriber_reward {
...
+  oneof entity {
+    // Subscriber reward
+    bytes subscriber_id = 1;
+    // Gateway reward
+    bytes gateway_key = 2;
+  }
+  // Amount in bones given by the service provider
+  uint64 service_provider_amount = 3;
+  // Amount in bones matched by the network
+  uint64 matched_amount = 4;
```
andymck, line 623: suggest renaming to network_amount

andymck: ignore, i see matched is the term used in the hip

bbalser (approved):

PR #415, opened 2024-08-27
add settlement status to data transfer session

michaeldjeffrey:
Oracles: https://github.com/helium/oracles/pull/857

- Added `data_transfer_session_req_v1.status` of enum `data_transfer_session_settlement_status`
  - indicates whether a session can have it's DC burned immediately, or needs manual intervention
- Added `pending_data_transfer_session_v1` for publicly auditable pending sessions

### DIFF
--- a/src/service/poc_mobile.proto
+++ b/src/service/poc_mobile.proto
@@ -711,8 +711,8 @@ message pending_data_transfer_session_v1 {
   uint64 rewardable_bytes = 5;
 
   // Timestamp in milliseconds since the epoch
-  // From the DataTransferEvent
-  // NOTE: Be sure to convert `data_transfer_event.timestamp` from seconds -> milliseconds
+  // From the DataTransferEvent, be sure to convert
+  // `data_transfer_event.timestamp` from seconds -> milliseconds
   uint64 event_timestamp = 6;
   // Timestamp in milliseconds since the epoch
   // From the file the transfer session was processed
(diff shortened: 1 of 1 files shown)


src/service/poc_mobile.proto
```diff
@@ -692,6 +700,24 @@ message invalid_data_transfer_ingest_report_v1 {
...
+  uint64 upload_bytes = 3;
+  uint64 download_bytes = 4;
+  uint64 rewardable_bytes = 5;
+
+  // Timestamp in seconds since the epoch
+  // From the DataTransferEvent
+  uint64 event_timestamp = 6;
+  // Timestamp in milliseconds since the epoch
+  // From the file the transfer session was processed
+  uint64 received_timestamp = 7;
```
kurotych, line 718: One timestamp in seconds, but another in milliseconds can be confusing.
Is it possible to use one format?

michaeldjeffrey: Agreed.
I was trying to respect the format of the message the timestamps originated from, but I'm not sure that adds anything beneficial. 

Moving to millis only.

bbalser (approved):

src/service/poc_mobile.proto
```diff
@@ -692,6 +700,25 @@ message invalid_data_transfer_ingest_report_v1 {
   uint64 timestamp = 3;
 }
 
+message pending_data_transfer_session_v1 {
+  // Public Key of the Hotspot
+  bytes pub_key = 1;
```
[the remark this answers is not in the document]
michaeldjeffrey: That's a good callout. 
They are named this way because most of the fields were lifted directly from `data_transfer_event`. 

Because of how tightly these two messages are related, I think it would be equally as confusing to have different field names between them.

PR #418, opened 2024-09-12
hip 118 visualization data 

andymck:
Adds usage count msgs to support hip 118 visualization data

### DIFF
--- a/src/service/poc_mobile.proto
+++ b/src/service/poc_mobile.proto
@@ -7,6 +7,67 @@ import "mapper.proto";
 import "hex_boosting.proto";
 import "service_provider.proto";
 
+message hex_usage_counts_req_v1 {
+  // H3 resolution 10 tile
+  uint64 hex = 1;
+  // Average count of unique service provider subscribers active in the hex
+  // during the previous 7 days ( inclusive of discovery mapping subscribers )
+  uint64 service_provider_subscriber_avg_count = 2;
+  // Average count of unique discovery mapping subscribers active
+  // in the hex during the previous 7 days
+  uint64 disco_mapping_avg_count = 3;
+  // Average count of unique offload sessions active in the hex during the
+  // previous 7 days
+  uint64 offload_avg_count = 4;
+  // Unix timestamp in milliseconds of when the report was generated
+  uint64 timestamp = 6;
+  // Carrier pubkey from verification mapping service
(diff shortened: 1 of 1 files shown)


src/service/poc_mobile.proto
```diff
@@ -7,6 +7,70 @@ import "mapper.proto";
...
 
+message hex_usage_counts_req_v1 {
+  // H3 resolution 10 tile
+  uint64 hex = 1;
+  // Average count of unique helium mobile subscribers active in the hex during
+  // the previous 7 days ( inclusive of discovery mapping subscribers )
+  uint64 helium_mobile_subscriber_avg_count = 2;
+  // Average count of unique helium mobile discovery mapping subscribers active
+  // in the hex during the previous 7 days
+  uint64 helium_mobile_disco_mapping_avg_count = 3;
```
jeffgrunewald, line 18: probably shouldn't make these helium-mobile specific, right? `carrier_subscriber_avg_count` and such?

src/service/poc_mobile.proto
```diff
@@ -7,6 +7,70 @@ import "mapper.proto";
...
+  // Average count of unique offload sessions active in the hex during the
+  // previous 7 days
+  uint64 offload_avg_count = 4;
+  // Average count of unique tmo cell tower sessions active in the hex during
+  // the previous 7 days
+  uint64 tmo_cell_avg_count = 5;
```
jeffgrunewald, line 24: assuming this is even a thing we're allowed to report (not restricted by any sort of service agreement) this should also probably be non-tmo specific; `mno_cell_avg_count` or similar

bbalser: agree with the naming. @madninja you verified that we are allowed to share this data?

bbalser: talked with ninja, lets rip out the tmo_cell_avg for right now

src/service/poc_mobile.proto
andymck: renamed to service_provider_subscriber_avg_count and disco_mapping_avg_count

src/service/poc_mobile.proto
andymck: removed

jeffgrunewald (approved):

bbalser (approved):

PR #419, opened 2024-09-12
Macpie/radio location estimates

macpie:

PR #426, opened 2024-10-25
Radio Location Estimates with hexes

michaeldjeffrey:
This is an extension of #419 that changes (lat,lon,radius) to a series of res12 hexes.


src/service/poc_mobile.proto
```diff
@@ -891,3 +893,48 @@ message verified_subscriber_verified_mapping_event_ingest_report_v1 {
   // Timestamp in milliseconds since unix epoch
   uint64 timestamp = 3;
 }
+
+message rle_event_v1 {
```
jeffgrunewald, line 897: what does this message represent?

jeffgrunewald (approved):

bbalser (approved):

src/service/poc_mobile.proto
michaeldjeffrey: Fantastic question!

Brian helped enlighten me that this message is series of correlations that leads to the location estimate. For data reason, the most we can expose of the correlation is the `id` and `timestamp`. 

It has been renamed `rle_event_v1` -> `radio_location_correlation_v1`.

jeffgrunewald: What does a correlation event entail? Some pure identifier feels like we’re outputting data simply for the sake of outputting data. I would think a safe thing to output that would also be intrinsically useful beyond token “transparency” would be to output an estimated location (in lat/lon or hex) with an estimated accuracy score and leave the info about the subscriber that provided that correlation the opaque portion that doesn’t escape our private backends

jeffgrunewald (approved):
Per my comment on the the correlation event message, I guess the `radio_location_estimate_v1` message type is doing exactly what I was asking for, so I guess my _real_ question is really "do the list of correlation events attached to an estimate really provide any value?" such that they belong in this public data structure. They have no meaning to anyone outside of the private backend that processed the location estimate so they just seem like noise here.

macpie (approved):

macpie (approved):

PR #428, opened 2024-11-07
Adds usage count msgs to support hip 118 visualization data

andymck:


src/service/poc_mobile.proto
```diff
@@ -7,6 +7,81 @@ import "mapper.proto";
 import "hex_boosting.proto";
 import "service_provider.proto";
 
+message hex_usage_counts_req_v1 {
```
dcyoung, line 10: I'm curious, do that data sources that drive hex usage include data transfer?  If so, would it be feasible to include the volume (bytes) of data transfer in a given hex as part of this payload? We currently aggregate data transfer leveraging hotspot locations - wondering if there could be a better way.

image

src/service/poc_mobile.proto
```diff
@@ -7,6 +7,81 @@ import "mapper.proto";
...
+  // H3 resolution 10 tile
+  uint64 hex = 1;
+  // Number of unique service provider subscribers active in the hex
+  // during the epoch ( inclusive of discovery mapping enabled subscribers )
+  uint64 service_provider_subscriber_count = 2;
+  // Number of unique discovery mapping enabled subscribers active
+  // in the hex during the epoch
+  uint64 disco_mapping_count = 3;
+  // Number of unique offload users active in the hex during the epoch
+  uint64 offload_count = 4;
```
dcyoung, line 20: if i understand correctly, these 3 fields represent quantities of users in various categories. Consider a common naming convention between the fields to indicate this... ie: `xxxx_user_count` or `xxx_subscriber_count`

src/service/poc_mobile.proto
dcyoung, line 10: Is the `counts` proto name conforming to an existing convention? I could imagine this payload growing to include usage details that are not strictly counts. Total nit, but names like `details` or `stats` may be more generic.  

Same applies to the radio counts message.

src/service/poc_mobile.proto
```diff
@@ -7,6 +7,81 @@ import "mapper.proto";
...
+  uint64 received_timestamp = 1;
+  // the verified report
+  hex_usage_counts_req_v1 report = 2;
+}
+
+message radio_usage_counts_req_v1 {
+  // The onchain address of the hotspot
+  bytes hotspot_pubkey = 1;
+  // the cbsd id of the radio if it is a cbrs radio otherwise empty
+  string cbsd_id = 2;
```
dcyoung, line 46: For scenario with multiple CBRS radios belonging to the same hotspot... will there exist multiple `radio_usage_counts..` events with the same `hotspot_pubkey` ?

src/service/poc_mobile.proto
```diff
@@ -7,6 +7,81 @@ import "mapper.proto";
...
+  // Number of unique service provider subscribers which have connected
+  // to the radio during the epoch ( inclusive of discovery mapping
+  // enabled subscribers )
+  uint64 service_provider_subscriber_count = 3;
+  // Number of unique discovery mapping subscribers which
+  // have connected to the radio during the epoch
+  uint64 disco_mapping_count = 4;
+  // Number of unique offload users which have connected to the radio
+  // during the epoch
+  uint64 offload_count = 5;
```
dcyoung, line 56: same as above - consider more common naming pattern to indicate user counts

src/service/poc_mobile.proto
```diff
@@ -7,6 +7,81 @@ import "mapper.proto";
...
+  // total count of bytes transferred via the radio by service provider
+  // subscribers during the epoch, inclusive of upload and download
+  uint64 service_provider_transfer_bytes = 6;
+  // total count of bytes transferred via the radio by offload users
+  // during the epoch
+  // inclusive of upload and download
+  uint64 offload_transfer_bytes = 7;
```
dcyoung, line 63: awesome!!!!

src/service/poc_mobile.proto
andymck: It is certainly possible to include the volume in bytes of data transfer on a per hex basis.  I understood you only wanted this per radio, I can include per hex also

src/service/poc_mobile.proto
andymck: that is correct

src/service/poc_mobile.proto
andymck: added here: https://github.com/helium/proto/pull/428/commits/238ac3e532215b35f719daba3b9fa59079318388

src/service/poc_mobile.proto
andymck: addressed here: https://github.com/helium/proto/pull/428/commits/1b9d9548a6f7c4d390cf0585b2d7c61f564e0d8a

src/service/poc_mobile.proto
andymck: addressed here: https://github.com/helium/proto/pull/428/commits/1b9d9548a6f7c4d390cf0585b2d7c61f564e0d8a

dcyoung:
@andymck looks great to me. I'm trying to think through our consumption downstream, and wondering about idempotency. We'll likely be ingesting events from a bucket populated by packet verifier, but wondering if there is a convention for identifying individual messages. Is a Radio ID + timestamp combo guaranteed to be unique?

dcyoung (approved):

dcyoung (approved):

bbalser (approved):

PR #430, opened 2024-11-18
Add deployment info to gateway_metadata

kurotych:


src/service/mobile_config.proto
```diff
@@ -21,10 +21,35 @@ import "reward_manifest.proto";
...
+  uint32 azimuth = 3;
+  uint32 mechanical_down_tilt = 4;
+  uint32 electrical_down_tilt = 5;
+}
+
+message cbrs_deployment_info {
+  repeated radio_deployment_info radio_deployment_info = 1;
+}
+
+message radio_deployment_info {
```
madninja, line 38: Prepend CBRS to this message?

kurotych: It makes sense, I will

jeffgrunewald (approved):

src/service/mobile_config.proto
```diff
@@ -21,10 +21,35 @@ import "reward_manifest.proto";
 // - Keypair fields are binary encoded public keys, Rust encoding example here:
 //    https://github.com/helium/helium-crypto-rs/blob/main/src/public_key.rs#L347-L354
 
+message wifi_deployment_info {
+  // antenna id
+  uint32 antenna = 1;
+  // The height of the hotspot above ground level in whole meters
+  uint32 elevation = 2;
```
andymck, line 28: maybe add comments for each field here providing some context of each field

andymck (approved):

src/service/mobile_config.proto
kurotych: I was thinking about that but nothing comes to mind except obvious: "Antenna azimuth" (etc)

PR #432, opened 2024-11-22
Add gateway_info_v2, gateway_info_stream_req_v2, gateway_metadata_v2

kurotych:
Since changes in gateway info response (#429, #430) breaks backward compatibility (due to signing message from server side and verifying on client):
- revert (v1) changes
- make v2 version
- make v1 version deprecated

### DIFF
--- a/src/service/mobile_config.proto
+++ b/src/service/mobile_config.proto
@@ -158,6 +158,16 @@ message gateway_info_stream_res_v1 {
   bytes signature = 4;
 }
 
+message gateway_info_stream_res_v2 {
+  // a list of gateway info numbering up to the request batch size
+  repeated gateway_info_v2 gateways = 1;
+  // unix epoch timestamp in seconds
+  uint64 timestamp = 2;
+  // pubkey binary of the signing keypair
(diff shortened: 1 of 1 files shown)


src/service/mobile_config.proto
```diff
@@ -43,6 +43,13 @@ message cbrs_radio_deployment_info {
 }
 
 message gateway_metadata {
+  option deprecated = true;
```
jeffgrunewald, line 46: i've only ever tried to deprecate fields with this notation, not entire messages but in those cases rust/prost seems to be unable to handle the message being deprecated and omitted without throwing missing field warnings. hopefully it's easier when the whole message is flagged

jeffgrunewald (approved):

src/service/mobile_config.proto
kurotych: Honestly, I’ve never thought about how Rust/Prost handles this deprecation, but I didn’t see any related warnings while working on this PR.

michaeldjeffrey (approved):

PR #434, opened 2024-12-06
Change `min_refreshed_at` to `min_updated_at` in `gateway_info_v2`

kurotych:
Related PR: https://github.com/helium/oracles/pull/910

### DIFF
--- a/src/service/mobile_config.proto
+++ b/src/service/mobile_config.proto
@@ -76,8 +76,9 @@ message gateway_info_v2 {
   gateway_metadata_v2 metadata = 2;
   // the asserted device type of the gateway
   device_type device_type = 3;
-  // The unix epoch timestamp (in seconds)
-  // when the data was read from chain and written to the db
+  // The Unix epoch timestamp (in seconds)
+  // when the data was read from the chain and written to the database.
+  // It does not necessarily mean that anything was changed in the gateway.
   uint64 refreshed_at = 4;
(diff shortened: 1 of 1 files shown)


src/service/mobile_config.proto
```diff
@@ -76,8 +76,9 @@ message gateway_info_v2 {
   gateway_metadata_v2 metadata = 2;
   // the asserted device type of the gateway
   device_type device_type = 3;
-  // The unix epoch timestamp (in seconds)
-  // when the data was read from chain and written to the db
+  // The Unix epoch timestamp (in seconds)
+  // when the data was read from the chain and written to the database.
+  // It does not necessarily mean that anything was changed in the gateway.
```
jeffgrunewald, line 81: is there any value in returning this field in the gateway info? seems to not really have much use for a caller of this api and a safe time to remove it if nothing is yet consuming the v2 messages

kurotych: > seems to not really have much use for a caller of this api

Agree. I've removed refreshed_at. I decided not to delete created_at, as it is the only "timestamp" field that provides an approximate date for the appearance of a hotspot on the Solana blockchain.

dcyoung:
@kurotych i noticed the v2 endpoint only applies to the stream endpoint. Should the `info` or `info_batch` endpoints get a v2 as well?

bbalser (approved):

PR #435, opened 2024-12-06
Feat/675 on chain ouis devaddrs

bryzettler:
Oui and Devadders are moving onchain. 

All write commands will be interacting directly with the solana blockchain via a solana_client now.

### DIFF
--- a/src/service/iot_config.proto
+++ b/src/service/iot_config.proto
@@ -18,7 +18,11 @@ import "region.proto";
 //   are signed by the config service to allow the recipient to validate
 //   the authenticity of the data returned.
 //
-// - Every key called `owner`, and `delegate_keys` are solana pubkeys
+// - org_v1 Every key called `owner`, `payer` and `delegate_keys` are binary
+//    encoded public keys, Rust encoding example here:
+//    https://github.com/helium/helium-crypto-rs/blob/main/src/public_key.rs#L347-L354
+//
+// - org_v2 Every key called `owner`, and `delegate_keys` are solana pubkeys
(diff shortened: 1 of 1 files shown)


macpie (approved):

src/service/iot_config.proto
```diff
@@ -43,14 +42,18 @@ enum action_v1 {
 // Define an organisation
 message org_v1 {
   uint64 oui = 1;
+  // On chain address
+  bytes address = 2;
```
jeffgrunewald, line 46: we can't change this proto like this without breaking any existing clients. we can add fields and that should be relatively safe for the old stuff to handle but changing the order of existing fields or adding fields in order ahead of existing fields will need an `org_v2` message to avoid a breaking change. unless we want to consider updating the "official cli" and the client in the iot oracles the only priority and let any other consumers deal with the fallout.

maybe that's acceptable in this special case since the shift to solana for apis using this particular data structure are going to be broken without an update no matter what ?

macpie: Good point actually, we need to decide if we want to v2 or nuke v1? I think we would be broken no matter what but maybe we v2, deprecate v1 and remove it at a later date.

jeffgrunewald: Only reason I don’t think it’s straightforward is b/c there are still methods that return that data structure that by the proto spec would be expected to still work

bryzettler: Yeah the data structure is going to change completely once we do the shift. I can do the v2 and deprecate v1 if thats the consensus?

macpie: Seems fair to me

bryzettler: Think I deprecated it properly, lemme know if thats how we want to do it

PR #441, opened 2025-02-12
Fix Rust CI build

kurotych:
set an upper bound on supported rand versions

### DIFF
--- a/beacon/Cargo.toml
+++ b/beacon/Cargo.toml
@@ -10,7 +10,7 @@ base64 = ">=0.21"
 byteorder = "1"
 helium-proto = { path = "../", features = ["services"] }
 prost = {  workspace = true }
-rand = "<=0.8.5"
+rand = "< 0.9"
 rand_chacha = "<=0.3.1"
 serde = { workspace = true }
 sha2 = "0"
--- a/beacon/Cargo.toml
+++ b/beacon/Cargo.toml
@@ -10,8 +10,8 @@ base64 = ">=0.21"
 byteorder = "1"
 helium-proto = { path = "../", features = ["services"] }
 prost = {  workspace = true }
-rand = ">=0.9"
-rand_chacha = ">=0.9"
+rand = "<=0.8.5"
+rand_chacha = "<=0.3.1"
 serde = { workspace = true }
 sha2 = "0"
 rust_decimal = { version =  "1", features = ["serde-with-float"] }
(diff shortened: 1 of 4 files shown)


andymck (approved):
LGTM, the alternative is to pin the deps at 0.8.  Might be less risk giving this gets pulled in by just about everything...lets see what @madninja thinks

beacon/Cargo.toml
```diff
@@ -10,8 +10,8 @@ base64 = ">=0.21"
 byteorder = "1"
 helium-proto = { path = "../", features = ["services"] }
 prost = {  workspace = true }
-rand = ">=0.8"
-rand_chacha = ">=0.3"
+rand = ">=0.9"
```
madninja, line 13: Would 0.8 work as lowest version?

kurotych: In case you want to explicitly set rand = 0.8 for this dependency—I think not. The breaking changes arrived in 0.9

But, I've tested the oracles repo with this branch. It has 3 different rand version inside and it compiles successfully 

```
oracles on  use-new-rand [?] via 🦀 v1.83.0 on ☁️  akurotych@nova-labs.com took 4s
❯ cargo tree -i rand
error: There are multiple `rand` packages in your project, and the specification `rand` is ambiguous.
Please re-run this command with one of the following specifications:
  rand@0.7.3
  rand@0.8.5
  rand@0.9.0
```

madninja: I meant >= 0.8

kurotych: > I meant >= 0.8 but that would break because it picks up 0.9, right?

Yes, correct

> I usually try to avoid many versions of crates. It works but bloats the binary size.
The usual way to fix this is with -
...
=0.8 to avoid yet another copy

Pining version to =0.8 is the another way to fix this problem. I don't have any strong opinion which is better.
Do you prefer pining the version to 0.8?

madninja: Yes please. Reducing duplicates is good

kurotych: > Yes please. Reducing duplicates is good

Done. I've reverted code changes and set an upper bound on supported rand versions to give Cargo more freedom in choosing the rand version.

beacon/Cargo.toml
```diff
@@ -10,8 +10,8 @@ base64 = ">=0.21"
 byteorder = "1"
 helium-proto = { path = "../", features = ["services"] }
 prost = {  workspace = true }
-rand = ">=0.8"
-rand_chacha = ">=0.3"
+rand = "<=0.8.5"
```
madninja, line 13: That’s too tight. =0.8

andymck: @madninja so you want it pinned at 0.8 ?

madninja: < 0.9

andymck (approved):

PR #442, opened 2025-03-11
Upgrade tonic and related crates

madninja:


madninja:
Caused issues talking to tonic services running at old versions. Shelving for now

PR #443, opened 2025-03-17
unify subscriber mapping rewardable activity submissions

jeffgrunewald:


src/service/poc_mobile.proto
```diff
@@ -985,3 +987,41 @@ message verified_subscriber_verified_mapping_event_ingest_report_v1 {
   // Timestamp in milliseconds since unix epoch
   uint64 timestamp = 3;
 }
+
+enum subscriber_mapping_type {
+  mapping_discovery = 0;
+  mapping_verification = 1;
+}
```
michaeldjeffrey, line 994: When you prefix an enum member with the enum name, proto will elide it.

```
enum subscriber_mapping_type {
  subscriber_mapping_type_discovery = 0;
  subscriber_mapping_type_verification = 1;
}
```

Will give you a rust generated type of
```
struct SubscriberMappingType {
  Discovery,
  Verification
}
```

madninja: oh that's neat! Allows proto to live in it's miserable flat namespace while the rust struct is clean

jeffgrunewald (commented):
👍

src/service/poc_mobile.proto
jeffgrunewald: nice to keep in mind for future reference but no longer needed for this change as we're unifying to a single mapping activity report with no need for a type-enum

madninja (approved):

PR #444, opened 2025-03-21
Feat/675 on chain ouis devaddrs

bryzettler:


src/service/iot_config.proto
```diff
@@ -185,16 +233,22 @@ message org_create_roamer_req_v1 {
...
 message org_update_req_v1 {
   message delegate_key_update_v1 {
+    option deprecated = true;
+
     bytes delegate_key = 1;
     action_v1 action = 2;
   }
 
   message devaddr_constraint_update_v1 {
+    option deprecated = true;
```
michaeldjeffrey, line 243: Can we get a comment of what replaced this since there's no `v2` replacement in the file?

michaeldjeffrey (approved):

PR #445, opened 2025-03-28
Add burn_timestamp to valid_data_transfer_session

macpie:
See https://github.com/helium/oracles/pull/980

### DIFF
--- a/src/service/packet_verifier.proto
+++ b/src/service/packet_verifier.proto
@@ -34,4 +34,5 @@ message valid_data_transfer_session {
   // session in
   uint64 last_timestamp = 7;
   uint64 rewardable_bytes = 8;
+  uint64 burn_timestamp = 9;
 }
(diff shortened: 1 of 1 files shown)


src/service/packet_verifier.proto
```diff
@@ -34,4 +34,5 @@ message valid_data_transfer_session {
   // session in
   uint64 last_timestamp = 7;
   uint64 rewardable_bytes = 8;
+  uint64 burn_timestamp = 9;
```
bbalser, line 37: does this field store a timestamp in seconds or milliseconds?

macpie: Good question! 2x checking

macpie: `Timestamp in millis dated when burn transaction is confirmed`

bbalser (approved):

michaeldjeffrey (approved):

PR #446, opened 2025-04-02
Hotspot Banning

michaeldjeffrey:
Supports: https://github.com/helium/oracles/pull/984

Add hotspot banning rpc and messages for `poc_mobile` ingester.

What are some more applicable reasons for ` ban_reason`?

### DIFF
--- a/beacon/src/region.rs
+++ b/beacon/src/region.rs
@@ -78,7 +78,7 @@ impl Region {
     pub fn from_i32(v: i32) -> Result<Self> {
         ProtoRegion::try_from(v)
             .map(Self)
-            .map_err(|_| Error::UnsupportedRegion(v))
+            .map_err(|_| Error::unsupported_region(v))
     }
 
     pub fn is_unknown(&self) -> bool {
--- a/beacon/src/region.rs
+++ b/beacon/src/region.rs
@@ -76,9 +76,9 @@ impl Default for Region {
 
 impl Region {
     pub fn from_i32(v: i32) -> Result<Self> {
-        ProtoRegion::from_i32(v)
+        ProtoRegion::try_from(v)
             .map(Self)
-            .ok_or_else(|| Error::unsupported_region(v))
+            .map_err(|_| Error::UnsupportedRegion(v))
     }
 
     pub fn is_unknown(&self) -> bool {
@@ -214,8 +214,8 @@ impl RegionParams {
             })
             // Convert to RegionSpreading
             .and_then(|region_spreading| {
-                RegionSpreading::from_i32(region_spreading)
-                    .ok_or_else(|| Error::unsupported_region_spreading(region_spreading))
+                RegionSpreading::try_from(region_spreading)
+                    .map_err(|_| Error::unsupported_region_spreading(region_spreading))
             })
     }
 
--- a/src/service/mobile_config.proto
+++ b/src/service/mobile_config.proto
@@ -213,6 +213,8 @@ enum admin_key_role {
   carrier = 3;
   // propagation calculation service of a mobile carrier
   pcs = 4;
+  // key for signing ban requests
+  admin_key_role_banning = 5;
 }
 
 enum network_key_role {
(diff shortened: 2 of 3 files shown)


beacon/src/region.rs
```diff
@@ -76,9 +76,9 @@ impl Default for Region {
 
 impl Region {
     pub fn from_i32(v: i32) -> Result<Self> {
-        ProtoRegion::from_i32(v)
+        ProtoRegion::try_from(v)
             .map(Self)
-            .ok_or_else(|| Error::unsupported_region(v))
```
madninja, line 81: Did the method on Error disappear? if so can we fix that?

src/service/mobile_config.proto
```diff
@@ -213,13 +213,17 @@ enum admin_key_role {
   carrier = 3;
   // propagation calculation service of a mobile carrier
   pcs = 4;
+  // key for signing ban requests
+  admin_key_role_banning = 5;
```
madninja, line 217: naming seams of? would it just not be `banning` to match he other names. or `admin_banning` to avoid the network enum collision?

src/service/poc_mobile.proto
```diff
@@ -394,6 +394,78 @@ service poc_mobile {
...
+  oneof ban_action {
+    ban_details_v1 ban = 7;
+    unban_details_v1 unban = 8;
+  }
+}
+
+message ban_details_v1 {
+  // Helpful to have for display purposes
+  string hotspot_serial = 2;
+  string notes = 3;
```
madninja, line 420: rename to `message`

src/service/poc_mobile.proto
```diff
@@ -394,6 +394,78 @@ service poc_mobile {
...
+  ban_type ban_type = 5;
+  // Timestamp in milliseconds that the ban should expire
+  // if not replaced by a subsequent ban.
+  // Note: 0 value means no expiration.
+  uint64 expiration_timestamp_ms = 6;
+}
+
+enum ban_reason {
+  ban_reason_unknown = 0;
+  ban_reason_gaming = 1;
```
madninja, line 431: ```ban_reason_location_gaming, ban_reason_data_farming```

src/service/poc_mobile.proto
```diff
@@ -394,6 +394,78 @@ service poc_mobile {
...
+}
+
+message unban_details_v1 {
+  // Helpful to have for display purposes
+  string hotspot_serial = 2;
+  string notes = 3;
```
madninja, line 437: message

beacon/src/region.rs
michaeldjeffrey: That is the method

src/service/mobile_config.proto
michaeldjeffrey: How do you feel about ?
```
admin_key_role.admin_banning;
network_key_role.network_banning;
```

beacon/src/region.rs
michaeldjeffrey: Whoops, I see now. Brought back

src/service/mobile_config.proto
michaeldjeffrey: Because of the proto enum name elision thing. the rust struct does read as `AdminKeyRole::Banning`.

madninja: I like clean rust names!

michaeldjeffrey: If we want to be able to use `AdminKeyRole::Banning` and `NetworkKeyRole::Banning` in rust, then yes.

madninja: Keep as is then

src/service/poc_mobile.proto
```diff
@@ -394,6 +394,79 @@ service poc_mobile {
...
+
+message ban_resp_v1 { uint64 timestamp_ms = 1; }
+
+message ban_req_v1 {
+  // Address of the hotspot being banned
+  bytes hotspot_pubkey = 1;
+  // Timestamp in milliseconds the message was sent
+  uint64 timestamp_ms = 3;
+  // Signer of the message
+  bytes ban_key = 5;
```
madninja, line 408: `ban_pubkey` to match `hotspot_pubkey`?

madninja (commented):
one more question/nit

madninja (approved):

PR #449, opened 2025-06-30
Add carrier_info to radio_usage_stats_req_v1

kurotych:


src/service/poc_mobile.proto
```diff
@@ -47,6 +47,25 @@ message hex_usage_stats_ingest_report_v1 {
...
+  carrier_0 = 0;
+  carrier_1 = 1;
+  carrier_2 = 2;
+  carrier_3 = 3;
+  carrier_4 = 4;
+  carrier_5 = 5;
+  carrier_6 = 6;
+  carrier_7 = 7;
+  carrier_8 = 8;
+  carrier_9 = 9;
```
madninja, line 60: there's no need to pre-allocate carriers.. we can add new ones later right?

bbalser (approved):

PR #451, opened 2025-07-09
[Mobile Config] new gateway service grpc endpoint: info_stream_v3

kurotych:
Related PR: https://github.com/helium/oracles/pull/1016. See description there

### DIFF
--- a/src/service/mobile_config.proto
+++ b/src/service/mobile_config.proto
@@ -85,9 +85,9 @@ message gateway_info_v2 {
 
 // Warning: Be careful, values in v2 differ from v1
 enum device_type_v2 {
-  indoor = 0;
-  outdoor = 1;
-  data_only = 2;
+  device_type_v2_indoor = 0;
+  device_type_v2_outdoor = 1;
+  device_type_v2_data_only = 2;
 }
 
 message location_info {
@@ -102,7 +102,7 @@ message deployment_info {
   uint32 antenna = 1;
   // The height of the hotspot above ground level, in whole meters
   uint32 elevation = 2;
-  // The azimuth (direction) of the antenna, in degrees
+  // The azimuth (direction) of the antenna, in degrees (whole number)
   uint32 azimuth = 3;
 }
 
--- a/src/service/mobile_config.proto
+++ b/src/service/mobile_config.proto
@@ -113,11 +113,12 @@ message gateway_info_v3 {
   // The Unix epoch timestamp (in seconds) when the location was last changed
   uint64 location_changed_at = 7;
 
-  // The Unix epoch timestamp (in seconds) when the gateway was first added to the database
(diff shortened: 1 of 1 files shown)


src/service/mobile_config.proto
```diff
@@ -83,6 +83,49 @@ message gateway_info_v2 {
...
+  uint64 location_changed_at = 2;
+}
+
+message deployment_info {
+  // The antenna ID of the gateway
+  uint32 antenna = 1;
+  // The height of the hotspot above ground level, in whole meters
+  uint32 elevation = 2;
+  // The azimuth (direction) of the antenna, in degrees
+  uint32 azimuth = 3;
```
michaeldjeffrey, line 106: Are we dealing with direction in whole units? 
Does this need to be a float, or are we treating it to some sort of decimal place?

src/service/mobile_config.proto
```diff
@@ -131,6 +174,27 @@ message gateway_info_stream_req_v1 {
   repeated device_type device_types = 4;
 }
 
+message gateway_info_stream_req_v3 {
+  // max number of gateway info records in each message of the response stream
+  uint32 batch_size = 1;
```
michaeldjeffrey, line 179: Is there a maximum batch size we're going to allow?

src/service/mobile_config.proto
kurotych: > Are we dealing with direction in whole units?

Yes

> Does this need to be a float

No, it doesn't. This field is always a whole number in the `mobile_hotspot_infos` table

src/service/mobile_config.proto
michaeldjeffrey: I was going to say something along the lines of, do we want to add a max batch_size as a heuristic so we can estimate the message length too large error without needing to hit the database and pull everything into memory first.

But then I did some digging.

https://github.com/hyperium/tonic/blob/v0.10.2/tonic/src/codec/mod.rs#L32-L34
```rs
// The default maximum uncompressed size in bytes for a message. Defaults to 4MB.
const DEFAULT_MAX_RECV_MESSAGE_SIZE: usize = 4 * 1024 * 1024;
const DEFAULT_MAX_SEND_MESSAGE_SIZE: usize = usize::MAX;
```

These settings are used for both Client and Server in tonic. (and they remain the same in the latest `0.13.1` version)

---
I tried a similar test, constructing the Client with a bigger max recv value. And the error wasn't thrown.
```rs
let mut client = GatewayClient::connect(addr)
        .await
        .unwrap()
        .max_decoding_message_size(20 * 1024 * 1024);
```

Client and Server both use the same encoding and decoding functions, weirdly for us, this means the request was successfuly, then the Client library generates the `Status` error.
https://github.com/hyperium/tonic/blob/v0.10.2/tonic/src/codec/decode.rs#L176-L188

PR #452, opened 2025-07-21
Mj/update tonic prost

michaeldjeffrey:
[Re-export tonic and prost as top level exports](https://github.com/helium/proto/commit/a6378d78bd3492407c64a72ad0b24ee9d8742d7d) 

This ensures downstream has access to all the types from the expected versions of tonic and prost. And clarifies where things like `DecodeError` come from in their imports.


michaeldjeffrey:
Closed in favor of https://github.com/helium/proto/pull/459

PR #455, opened 2025-08-12
chain rewardable entities

michaeldjeffrey:


src/service/chain_rewardable_entities.proto
```diff
@@ -0,0 +1,177 @@
...
+package helium.chain_rewardable_entities;
+
+enum pub_key_format {
+  pub_key_format_UNKNOWN = 0; // Unknown format, should not be used
+  pub_key_format_B58 = 1;     // Base58 format
+  pub_key_format_UTF8 = 2;    // UTF8 format
+}
+
+message pub_key {
+  string value = 1;
```
ChewingGlass, line 17: Not actually sure what this is representing? An entity key? A solana pubkey? String and not bytes here seems odd

src/service/chain_rewardable_entities.proto
```diff
@@ -0,0 +1,177 @@
...
+message mobile_hotspot_metadata {
+  // Timestamp when the NFT was created
+  uint64 created_at_ms = 1;
+  // Serial number of the mobile hotspot device (can be null or empty... in the
+  // case of data-only devices)
+  optional string serial_number = 2;
+  // Type of the mobile hotspot device
+  mobile_hotspot_device_type device_type = 3;
+  // The asserted location in hex format
+  string asserted_hex = 4;
```
ChewingGlass, line 41: Do we not want the other stuff from mobile hotspot info? Elevation? Device tilt? Etc?

src/service/chain_rewardable_entities.proto
```diff
@@ -0,0 +1,177 @@
...
+  string asserted_hex = 2;
+}
+
+// ------------------------------------------------------------------
+// Rewardable Entity <---> Controlling Wallet
+// ------------------------------------------------------------------
+enum re_controller_type {
+  re_controller_type_UNKNOWN = 0; // Unknown control type, should not be used
+  re_controller_type_DIRECT_OWNER = 1; // Owner of the NFT
+  re_controller_type_WELCOME_PACK_CONTROLLER =
```
ChewingGlass, line 59: Missing one here for fanout type. It is either in a welcome pack, or in a fanout. Not both at the same time.

src/service/chain_rewardable_entities.proto
```diff
@@ -0,0 +1,177 @@
...
+}
+
+message mobile_hotspot_change_resp_v1 { uint64 timestamp_ms = 1; }
+
+// The report which ingestor/oracles will WRITE to s3
+message mobile_hotspot_change_report_v1 {
+  // The time the server received the request
+  uint64 received_timestamp_ms = 1;
+  // The request report
+  mobile_hotspot_change_req_v1 report = 2;
```
ChewingGlass, line 133: Why do we have this extra nesting?

ChewingGlass (changes requested):

src/service/chain_rewardable_entities.proto
dcyoung: MVP use cases don't need it - but could well be useful to others. Hopefully others can offer some perspective here - @madninja or maybe even someone from HG

src/service/chain_rewardable_entities.proto
```diff
@@ -0,0 +1,203 @@
...
+}
+
+message rewards_split_v1 {
+  // They key of the on-chain struct that contains/performs the split
+  solana_pub_key pub_key = 1;
+  // The cron schedule of the split
+  string schedule = 2;
+  // The total shares of the split (user can specify any number of shares,
+  // defaults to 100)
+  uint32 total_shares = 3;
```
michaeldjeffrey, line 78: This is a question for my own curiosity. 

How do we ensure `split_recipient_info_v1` shares don't exceed `total_shares`?

What happens if `split_recipient_info_v1` have fixed amount of rewards, but not enough rewards are issued in a given epoch?

ChewingGlass (approved):

dcyoung (approved):

PR #456, opened 2025-08-13
Update rewardable entities protos

ChewingGlass:


src/service/chain_rewardable_entities.proto
```diff
@@ -24,30 +21,36 @@ message pub_key {
...
   mobile_hotspot_device_type device_type = 3;
   // The asserted location in hex format
-  string asserted_hex = 4;
+  optional string asserted_hex = 4;
+  optional uint32 antenna = 5;
+  optional int32 elevation = 6;
+  optional uint32 azimuth = 7;
+  optional uint32 mechanical_down_tilt = 8;
+  optional uint32 electrical_down_tilt = 9;
+  bool is_full_hotspot = 10;
```
dcyoung, line 43: what does `is_full_hotspot` mean? 

Are these fields optional for the case where an NFT exists, but a hotspot has not yet been asserted? 

We might consider NOT emitting events until the hotspot has been asserted.

src/service/chain_rewardable_entities.proto
```diff
@@ -24,30 +21,36 @@ message pub_key {
...
-  mobile_hotspot_device_type_WIFI_DATA_ONLY = 3;
+  mobile_hotspot_device_type_CBRS = 1;
+  mobile_hotspot_device_type_WIFI_INDOOR = 2;
+  mobile_hotspot_device_type_WIFI_OUTDOOR = 3;
+  mobile_hotspot_device_type_WIFI_DATA_ONLY = 4;
 }
 
 message mobile_hotspot_metadata {
-  // Timestamp when the NFT was created
-  uint64 created_at_ms = 1;
```
dcyoung, line 34: would like to hold onto this - it's our "deployed_at" marker for downstream ETL. Is there perhaps a better place to put it?

src/service/chain_rewardable_entities.proto
```diff
@@ -62,38 +65,39 @@ enum re_controller_type {
...
+  oneof reward_amount {
+    // Fixed amount of tokens to be distributed to the wallet each period.
+    uint64 fixed_amount = 4;
+    // The shares for this wallet in the fanout. Divide into split.total_shares for percentage.
+    uint32 shares = 5;
+  }
+  // If the user has closed their token account, we can't transfer tokens to them. This is the amount of tokens thay are owed
+  // that will distribute to them in the next payment if they recreate the ATA.
+  optional uint64 pending_payments = 6;
 }
```
dcyoung, line 97: This seems like creeping implementation detail into the payload. Already reasonably complex to interpret. At the end of the day, the MVP need is the "downstream user wallet + reward share + schedule" for any direct or indirect recipients.   

If you think this too limiting, how about a minimal subset of this change:

```proto
message split_v1_recipient_info {
  // The identifier for the user wallet that receives rewards
  solana_pub_key recipient_wallet = 1;
  oneof reward_amount {
    // Fixed amount of tokens to be distributed to the wallet each period.
    uint64 fixed_amount = 2;
    // The shares for this wallet in the fanout. Divide into split.total_shares for percentage.
    uint32 shares = 3;
  }
}

message rewards_split_v1 {
  // the pub-key of the fanout wallet which conducts the split
  solana_pub_key fan_out_pub_key = 1;
  
  string cron_schedule = 2;
  // total shares  (Noah, can you elaborate on what this means here... is this always 100?)
  uint32 total_shares = 3;
  repeated split_v1_recipient_info recipients = 4;
}
```

```
message atomic_hotspot.... {
  ...
  oneof rewards_destination {
    solana_pub_key recipient_wallet = 7;
    rewards_split_v1 rewards_split_v1 = 8;
  }
}
```

src/service/chain_rewardable_entities.proto
```diff
@@ -104,33 +108,38 @@ message re_reward_recipient_info {
...
-  mobile_hotspot_metadata metadata = 3;
+  mobile_hotspot_metadata metadata = 4;
 
   // Controlling ownership of the NFT
-  re_controller_info controlling_ownership = 4;
+  re_controller_info owner = 5;
 
-  // Reward recipients for the NFT
-  repeated re_reward_recipient_info recipients = 5;
+  uint64 lifetime_claimed_rewards = 6;
```
dcyoung, line 127: can drop this - feels more like "stat" information about the behavior of a hotspot.

src/service/chain_rewardable_entities.proto
ChewingGlass: You can actually assert a hotspot without this information. Deployment info is optional.

is_full_hotspot is how we denote data only hotspots. Basically if it's eligible for PoC. We can rename the field, but this is what we use everywhere else iirc

src/service/chain_rewardable_entities.proto
ChewingGlass: I'm erring on the side of adding more information because you might end up wanting this stuff to debug later on. 

For example, if `pending_payments` is nonzero, you may want to prompt the user to create a token account. If a use complains their tasks aren't running, you have task IDs to debug. You can also use those to display the next task run time. 

On the shares front -- the creator of the fanout can decide how many shares. 100. 1000. Etc. I don't like using something like a double here, where potentially you lose information due to rounding. Better to just pass through the total number of shares and divide when you're doing the ETL, if you want it that way.

src/service/chain_rewardable_entities.proto
ChewingGlass: Erring on the side of more information -- you will want this in the future, and you're not going to have a great way to get it if we're not publishing it here.

src/service/chain_rewardable_entities.proto
dcyoung: is_full_hotspot is a bit cryptic and sounds like its already covered by `device_type`. If it overlaps with device_type, let's remove it.

src/service/chain_rewardable_entities.proto
dcyoung: The shares stuff makes sense. We can enforce standard through UX, but that keeps it flexible and in full integer units.

The debugging you describe sounds like implementation detail though - appropriate for the internals of these systems, which is likely where someone would be performing said debugging yeah?

src/service/chain_rewardable_entities.proto
dcyoung: Hmm i think we've been using this as the marker for `deployed_at`: https://github.com/helium/proto/blob/master/src/service/mobile_config.proto#L80

Which seems potentially broken. Downstream ETL can likely maintain this in app DB and backfill with the current values we have. Though interesting that the existing values we've been using may not actually represent a deployment time.

